PR #9, opened 2019-09-19
PEN-315 oEmbeds Implementation without Parsing/Sanitizing

shk172:
[JIRA Ticket](https://arcpublishing.atlassian.net/browse/PEN-315)
More details on the comments on the ticket

Passed existing unit test

Note: As I've commented on the ticket: Sanitization would still be a problem but I feel that that should be done on Content API’s side if need be instead of bringing it to frontend on Pagebuilder.


brentmiller1973 (approved):
Awesome!

PR #20, opened 2019-10-15
PEN-397 Add better server-side render for Date block

shk172:
[JIRA Ticket](https://arcpublishing.atlassian.net/browse/PEN-397)
Because the full-icu library is too big (50mb uncompressed), we'll scrap the internationalization of the dates for now. Instead, we can leverage the `Intl` object and just use English locale at all times, which will match the client-side formatting. This will obviously be an issue for international customers if they wish to use their local time formatting - will have to address that in the future.

Also dependent on adding an additional timezone env variable on Maestro so that the customers can set their own timezone if they prefer. Otherwise it'll use default timezone on the server before showing (and flickering) the client's timezone retrieved from the browser

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -11,6 +11,7 @@ module.exports = {
             'fusion:themes': './jest/mocks/themes.js',
             'fusion:context': './jest/mocks/context.js',
             'fusion:consumer': './jest/mocks/consumer.js',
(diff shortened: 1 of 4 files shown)


jgrosspietsch (approved):
:cool:

PR #22, opened 2019-10-16
PEN-349 Switch designs for pullquote and blockquote

shk172:
[JIRA Ticket](https://arcpublishing.atlassian.net/browse/PEN-349)
Also forgot to do rest of the PEN-350 (only fixed it for blockquote, not pullquote) so doing that here too


jgrosspietsch (approved):
:cool:

PR #35, opened 2019-11-01
PEN-413 Convert Article Body Feature into a Chain

shk172:
[JIRA Ticket ](https://arcpublishing.atlassian.net/browse/PEN-413)
The only major change in terms of function is the lines 148-209 of the `blocks/article-body-block/chains/article-body-chain.jsx` file. The rest is just copy from the children of the article body block

Just need to add testing for it, but for now the main code changes can be reviewed in the meanwhile


jgrosspietsch (approved):
:cool:

PR #40, opened 2019-12-01
Pen 495

brentmiller1973:
There are some discrepancies in the design that are inconsistent (colors, fonts, spacing, etc).  When those are flushed out through a QA design, a second pass will be done.


jgrosspietsch (approved):
:cool:

PR #42, opened 2019-12-05
PEN-528 Unit tests for article body

RohitTWP:
Unit tests for article-body chain.
Ignored article-body feature from istanbul test coverage in order to commit my changes and create a pr.
Ignored unit tests for amp code on oembed as two of the lines are not behaving as expected.


jgrosspietsch (changes requested):
I have a bunch of mostly semantic change suggestions. Some are more organizational.

PR #69, opened 2019-12-26
Results list block main

RohitTWP:
Results List Block for News Theme


brentmiller1973 (commented):
I think we should take another look at this.  I think if we should be doing the conditional style logic with-in the scope of styled components (See: https://stackoverflow.com/questions/48502647/conditional-rendering-in-styled-components)

jgrosspietsch (approved):
:cool:

beltrancaliz (approved):

PR #70, opened 2020-01-02
Simple results list

RohitTWP:


jgrosspietsch (changes requested):

beltrancaliz (approved):

beltrancaliz (approved):

brentmiller1973 (commented):
Looks good, but images coming from the content source needs to leverage the Image component in the engine-sdk so that it can get resized.

PR #90, opened 2020-01-28
PEN-485: Quick fix for author content source block

beltrancaliz:
Implemented a fix for the author API content source block


jgrosspietsch (approved):
:cool:

PR #95, opened 2020-01-29
PEN-505: Design QA

RohitTWP:
1. As per the feedback from design QA the images should be 25% of the list item.
2. Explicitly set the height of the image so that all the images in the list have the same height.

https://arcpublishing.atlassian.net/browse/PEN-505


jgrosspietsch (approved):
:cool:

beltrancaliz (approved):

PR #97, opened 2020-01-30
PEN-564: Overline text and URL can be overriden by `label` in ANS doc

seans887:
JIRA ticket: https://arcpublishing.atlassian.net/browse/PEN-564

This changes which text/URL combo the `Overline` block uses based on the structure of the ANS document it is based on. Previously it just displayed the text from `websites[site-name]. website_section.name` as the text of the `<a>` tag, and `websites[site-name]. website_section._id` as the URL for the href. Now, if the ANS property `label.basic.display` exists and is set to true, the text of the anchor text should be the value of `label.basic.text` and the href should be `label.basic.url`.

I added tests to make sure the name and URL are overriden properly when the `label.basic.display` is true, and when it is NOT true.

Here's an example ANS doc with `label.basic` data set properly to override the section name (obv needs creds to access): https://api.corecomponents.arcpublishing.com/content/v4/?website_url=/2019/09/17/article-with-a-long-headline-lorem-ipsum-dolor-sit-amet-consectetur-adipiscing-elit/&website=the-sun

### DIFF
--- a/blocks/overline-block/features/overline/default.jsx
+++ b/blocks/overline-block/features/overline/default.jsx
@@ -10,36 +10,38 @@ const StyledLink = styled.a`
   text-decoration: none;
 `;
 
-function _get (obj, ...args) {
-  return args.reduce((obj, level) => obj && obj[level], obj)
-}
-
 const Overline = () => {
-  const { globalContent: content, arcSite } = useFusionContext();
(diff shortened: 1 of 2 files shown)


jgrosspietsch (approved):
:cool:

PR #100, opened 2020-01-31
Fixed responsive layout issue

beltrancaliz:
Allows for responsive layout on iPad and devices with odd widths


jgrosspietsch (approved):
:cool:

PR #109, opened 2020-02-11
Pen 567 card list

brannytempy:
Design QA The Card List. Images currently get too tall when occupying a full-width space. Use object-fit cover and max-height behavior so we can control this a bit more. May need a better solution down the road.


jgrosspietsch (approved):
:cool:

JackHowa (approved):

PR #139, opened 2020-02-28
[PEN-631] GitHub package config

shk172:
don't merge until documentation is done


JackHowa (commented):
Nice work. Any readme process changes?

brentmiller1973 (approved):
Bet that was fun!  :-P

JackHowa (commented):
I think the simple results list shouldn't be in here. One more thing is in the .eslintrc there's a couple ignores that may have to be updated to ignore wp media

beltrancaliz (approved):

PR #201, opened 2020-04-06
PEN-473: Navigation Block Search

beltrancaliz:
Refactored search Results list to utilize global content and custom content config
Added ability for the navigation block to search content
[Jira ticket](https://arcpublishing.atlassian.net/browse/PEN-473)


brentmiller1973 (changes requested):
See the comment on the timeout.  Also check to see if your code is fully keyboard accessible....if its not, then it needs to be called out so that it can be fixed in a later ticket.

JackHowa (approved):
🆒 -> sounds like Brent's accessibility/perf comment is a future issue

PR #207, opened 2020-04-09
PEN-794: Additional websites

beltrancaliz:
Added Google Fonts to the default output type for the new demo websites
[Jira ticket](https://arcpublishing.atlassian.net/browse/PEN-794)


JackHowa (approved):
Nice work. This one probably wasn't the most fun

PR #224, opened 2020-04-16
[PEN-838] Use a media query instead of a resize listener

jgrosspietsch:
This gets rid of our need to do window width checks explicitly or run a function on every resize change. These changes also prevent a server-side issue where `window.innerWidth` breaks the render.


JackHowa (approved):
🆒 x2

brentmiller1973 (approved):

PR #228, opened 2020-04-17
change from alert bar source to collections one

shk172:


ghost (approved):
Matches api call on line 20. Looks good.

PR #290, opened 2020-05-18
PEN-937: Missing trailing slashes

beltrancaliz:
Made a function to add a trailing slash if missing one for the following blocks:
- Links Bar
- Header nav feature
- Header nav chain
- Overline

Would be good to make a directory for reusable functions like this and the `extractImage` and `constructHref` functions to avoid repeated code.

[Jira ticket](https://arcpublishing.atlassian.net/secure/RapidBoard.jspa?rapidView=545&projectKey=PEN&modal=detail&selectedIssue=PEN-937)


JackHowa (approved):
Doing the dirty work. nice getting through it

PR #306, opened 2020-05-26
[PEN-603] Added initial internationalized blocks

jgrosspietsch:
Added `fusion:intl` support for the blocks that need it. Also, created a `global-phrases-block` that includes just an `intl.json` with "global" translations like those needed for the gallery.


JackHowa (approved):
Good work -- solid documentation

beltrancaliz (approved):

PR #331, opened 2020-06-05
Canary build action

jgrosspietsch:
Lerna's `--canary` option doesn't really do what we need (packages versioned with SHAs so that they can be built without version considerations) so this is a slightly custom solution. I want there to be an automated release that either gets published on every push to staging or on a certain interval (maybe someday).


JackHowa (approved):
Think this should be gtg if we have the secrets setup

brentmiller1973 (approved):

PR #358, opened 2020-06-16
[PEN-931] Documentation for our block dist-tags

jgrosspietsch:
Added descriptions and explanations of our block dist-tags. Also added an NPM script for automatically keeping parity between `stable` and `latest` in case our Github Actions fail.


JackHowa (approved):
:cool:

beltrancaliz (approved):

PR #363, opened 2020-06-23
PEN-941 emit message gallery autoplay

kascote:
[PEN-941](https://arcpublishing.atlassian.net/browse/PEN-941)

- Update developer notes about events en general y Gallery events in particular

# How was this tested?

no needed

- none


brentmiller1973 (commented):
I’m a little confused here.  I thought I already implemented this....An event _is_ thrown when ever the gallery moves to the next or previous image.  If the image moves via autoplay, then the event will have the property autoplay: true, If it was moved manually (user interaction), the the event will have the property autoplay: false,Also all the other Acceptance criteria was covered…unless there needs to be additional docs added?

PR #371, opened 2020-06-30
[PEN-1122] Updated deployment process to document new workflow with canary blocks

jgrosspietsch:
Updated the deployment instructions for our new canary+non-independent versioning deployment setup.

I didn't really change the hotfix instructions because if I have to come up with another hypothetical build process for our blocks I _will_ cry.


JackHowa (approved):
Looks good Joe, nice work

PR #380, opened 2020-07-07
PEN-1128 gallery inherit global content behavior

kascote:
[PEN-1128](https://arcpublishing.atlassian.net/browse/PEN-1128)

- fix an render issue when the gallery was in a template with global content

# How was this tested?
- tests added/fixed
- visually in PB with pages and templates
- screen record added to the ticket

- none


JackHowa (approved):
this looks ok from a syntax perspective. have a few naming nits. thanks for including the video; I think that it has been validated by qa. in the future, is there a way to limit the spacing updates that are changed

PR #384, opened 2020-07-07
PEN-1033 defensive check website url

kascote:
[PEN-1033](https://arcpublishing.atlassian.net/browse/PEN-1033)

- defensive check when there is not content for the arcSite

Updated the following blocks so that they first check if a website_url value exists for the website being rendered, and if it does not exist, skip rendering that item and move on

- Top Table List
- Results List
- Card List
- Simple List
- Numbered List

# How was this tested?
- tests written for each block

- none


JackHowa (commented):
Please include the desired behavior in this pr or any screenshots. did not know the goal of this pr until I read the ticket

```

Update the following blocks so that they first check if a website_url value exists for the website being rendered, and if it does not exist, skip rendering that item and move on

Top Table List

Results List

Card List

Simple List

Numbered List

```

JackHowa (approved):

beltrancaliz (approved):

PR #387, opened 2020-07-09
Updated production release docs

jgrosspietsch:
This is to prevent issues when there are changes in staging since the last beta release

[PEN-](https://arcpublishing.atlassian.net/browse/PEN-)

-
-

# How was this tested?

Hopefully, unit tests were written.

Otherwise, explain how to test this change.

If the amount of work to write unit tests for this change are excessive,

# Show Effect Of Changes

## Before

When I clicked the search button, the button turned red.

[include screenshot or gif or link to video, storybook would be awesome]

## After

When I clicked the search button, the button turned green.

[include screenshot or gif or link to video, storybook would be awesome]

Examples of dependencies or side effects are:


JackHowa (approved):
🆒

beltrancaliz (approved):

PR #398, opened 2020-07-22
Added some optimizations for reducing canary build times

jgrosspietsch:
This does 3 things that should speed up our canary builds:
- Caching NPM files ([We're caching `~/.npm` instead of `node_modules` because that's what's recommended.](https://github.com/actions/cache/blob/main/examples.md#node---npm))
- `npm ci` instead of `npm install`
- Removing `--force-publish` from our publish command to reduce the number of packages being published.

No, I haven't tested this and I won't test this.

### DIFF
--- a/.github/workflows/canary-build.yml
+++ b/.github/workflows/canary-build.yml
@@ -14,14 +14,21 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
+      - name: Cache NPM files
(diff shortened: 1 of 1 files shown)


brentmiller1973 (approved):
Nice man...also like how "It is also more strict than a regular install, which can help catch errors or inconsistencies caused by the incrementally-installed local environments of most npm users."

beltrancaliz (approved):

PR #399, opened 2020-07-22
PEN-1196 rework blocks spacing

kascote:
[PEN-1196](https://arcpublishing.atlassian.net/browse/PEN-1196)

- rework the spacing between blocks to be handled by the container instead of individually by each block

# How was this tested?
- test updated with the changes
- visually with different pages and layouts

# Show Effect Of Changes
### each element in a chain is spaced with the same rules

- need css framework to be on this same branch name or updated with this PR https://github.com/WPMedia/news-theme-css/pull/21


JackHowa (approved):
🆒

PR #401, opened 2020-07-23
Pen 1197 image resizer handle missing urls

kascote:
[PEN-1197](https://arcpublishing.atlassian.net/browse/PEN-1197)

- when the promo_items data was incomplete, the procedure break server side, this fix add more control

# How was this tested?

- tests written

# Show Effect Of Changes

## After

- none


JackHowa (approved):
good job adding the tests

PR #411, opened 2020-07-31
PEN-1137: Public Block Packages

beltrancaliz:
[PEN-1137](https://arcpublishing.atlassian.net/browse/PEN-1137)

- Removes restricted publishing config
- Adds a license to the repo
- Adds CC-NC-ND license to each package

# How was this tested?

Will test by installing blocks without a `.npmrc` once the repo becomes public

- This PR will be required to make the blocks repo public. 
- Since we are not making other dependencies like engine theme SDK or news theme css public yet, we will rely on having Luke's personal Github token, but it isn't publicly readable.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,1 +1,10 @@
-UNLICENSED
+Shield: [![CC BY-NC-ND 4.0][cc-by-shield]][cc-by-nc-nd]
+
+This work is licensed under a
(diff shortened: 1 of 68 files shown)


jgrosspietsch (approved):
:cool:

JackHowa (approved):

PR #427, opened 2020-08-12
Add ans fields for relevant blocks

JackHowa:
- https://arcpublishing.atlassian.net/secure/RapidBoard.jspa?rapidView=545&projectKey=PEN&modal=detail&selectedIssue=PEN-1138&assignee=5e387d6a1b1d910e5dfd7a9b

- document ans fields

### DIFF
--- a/blocks/card-list-block/README.md
+++ b/blocks/card-list-block/README.md
@@ -21,7 +21,10 @@ _Please provide a 1-2 sentence description of what the block is and what it does
 Outline any schema information requirements necessary to know for ths block
 
 ### ANS Fields
-- `Add all ANS fields used in the block`
+- `content_elements[x].websites[arcSite].website_url`
+- `content_elements[0].credits.by` Check for multiple authors in first article
+- `content_elements[0].websites[arcSite]` Uses `website_url`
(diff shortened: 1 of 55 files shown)


blocks/card-list-block/README.md
```diff
@@ -21,7 +21,10 @@ _Please provide a 1-2 sentence description of what the block is and what it does
 Outline any schema information requirements necessary to know for ths block
 
 ### ANS Fields
-- `Add all ANS fields used in the block`
+- `content_elements[x].websites[arcSite].website_url`
```
beltrancaliz, line 24: Should this be in JSON?

JackHowa: could be. I was going off the template that Robin had setup for the ticket:

author bio readme
```
### ANS fields
- `credits.by`
```

RobinGiannattasio: So, I added this format because as a fusion developer, we always get the response back from content sources as a usable object. I thought it was easier to read/document which particular fields were being used by doing it this way. I personally think the readability is improved by having a simple list instead of writing out the JSON longhand, but I could be wrong!

PR #429, opened 2020-08-14
PEN-429 taboola widget

kascote:
[PEN-429](https://arcpublishing.atlassian.net/browse/PEN-429)

- implements a new ad block for the Taboola widget

# How was this tested?
- tests written
- visually on PB

# Show Effect Of Changes

## After
blocks rendered inside PB editor

working widgets on page preview

- Need to setup the `taboolaPublisherId` on blocks json and valid widget values on each widget on the page

### DIFF
--- a/blocks/ad-taboola-block/features/ad-taboola/default.jsx
+++ b/blocks/ad-taboola-block/features/ad-taboola/default.jsx
@@ -1,0 +1,162 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import getProperties from 'fusion:properties';
+import Consumer from 'fusion:consumer';
+
+const taboolaLoader = (publisherID) => (
+  `window._taboola = window._taboola || [];
+    _taboola.push({article:'auto'});
+    !function (e, f, u, i) {
+      if (!document.getElementById(i)){
+        e.async = 1;
+        e.src = u;
+        e.id = i;
+        f.parentNode.insertBefore(e, f);
+      }
+    }(document.createElement('script'),
+    document.getElementsByTagName('script')[0],
+    '//cdn.taboola.com/libtrc/${publisherID}/loader.js',
+    'tb_loader_script');
+    if(window.performance && typeof window.performance.mark == 'function')
+      {window.performance.mark('tbl_ic');}
+  `
+);
+
+const taboolaFlusher = () => (
+  `
+  window._taboola = window._taboola || [];
+  _taboola.push({flush: true});
+  `
+);
+
+@Consumer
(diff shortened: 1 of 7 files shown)


blocks/ad-taboola-block/features/ad-taboola/default.jsx
```diff
@@ -0,0 +1,162 @@
...
+    const flusher = document.getElementById('tbl-flusher');
+    if (flusher) {
+      return;
+    }
+    const body = document.getElementsByTagName('body')[0];
+    if (!body) {
+      return;
+    }
+
+    this.appendScript('tbl-flusher', body, taboolaFlusher);
```
visionbegin, line 95: Just wondering if this might need to check for window undefined before accessing the DOM?  In other words, when this initially runs SSR...is this ok?

blocks/ad-taboola-block/features/ad-taboola/default.jsx
```diff
@@ -0,0 +1,162 @@
...
+      <>
+        <div id={`${container}`} />
+        <hr />
+        <script type="text/javascript" dangerouslySetInnerHTML={{ __html: scriptString }} />
+      </>
+    );
+  }
+}
+
+AdTaboola.label = 'Taboola – Arc Block';
```
visionbegin, line 143: Is Taboola going to be the only ad service we offer?  I just feel 'Taboola – Arc Block' is not descriptive enough...maybe something like 'Taboola Ad Svc – Arc Block' ?

brentmiller1973 (commented):
Not familiar with Taboola.  However, have a couple concerns that I marked in various comments

1.  Do we envision having other ad provider types? If so, we need to ensure our naming of the ad blocks are descriptive and follow a pattern.
2.  Want to make sure that any DOM manipulation (with the exception of script tag generation is only occurring on the client side to prevent SSR errors.

blocks/ad-taboola-block/features/ad-taboola/default.jsx
kascote: The `appendScript` function is called throw `insertLoader` and `insertFlusher`, and both are called on `componentDidMount` that is only execute client side.

SSR just will render the main container inside the page, but only if all the parameters are setup (publisherID and block parameters)

blocks/ad-taboola-block/features/ad-taboola/default.jsx
kascote: I'll add the comment to the ticket to check with Sara, I think make sense.

brentmiller1973 (approved):

PR #430, opened 2020-08-17
fix GA script insertion

kascote:
- GA script was being inserted incorrectly
- add test to check when there are nested <script> tags on the rendered output block

# How was this tested?

- test written

# Show Effect Of Changes

## After

- none


JackHowa (approved):
Looks good -- that makes sense

PR #434, opened 2020-08-26
PEN-1251- hide/show nav bar logo when scroll

kascote:
[PEN-1251](https://arcpublishing.atlassian.net/browse/PEN-1251)

- hide the nav bar logo when there is a masthead block present.
- show the nav bar logo when the masthead scroll off the screen

# How was this tested?
- no test at this point. the implementation use timers and scroll events
- was tested visually on PB and preview pages on all different viewports

# Show Effect Of Changes

## Before
on page load both logos are present

## After
pen-1251-1

on mobile is disabled, and the logo is always present
pen-1251-2

- none


blocks/header-nav-chain-block/chains/header-nav-chain-block/useDebounce.js
```diff
@@ -0,0 +1,84 @@
+// https://github.com/xnimorz/use-debounce
+import { useRef, useCallback, useEffect } from 'react';
+
+export default function useDebouncedCallback(callback, delay, options = {}) {
+  const { maxWait, leading, trailing = true } = options;
+  const maxWaitHandler = useRef(null);
+  const maxWaitArgs = useRef([]);
+
+  // const trailing = options.trailing === undefined ? true : options.trailing;
```
beltrancaliz, line 9: Was this comment left here intentionally?

beltrancaliz:
@kascote can we add some new unit tests to cover this functionality?

blocks/header-nav-chain-block/chains/header-nav-chain-block/useDebounce.js
```diff
@@ -0,0 +1,84 @@
+// https://github.com/xnimorz/use-debounce
```
JackHowa, line 1: can we use useDebounce package rather than copy-paste their code? if we use the package as a dependency, then we can get updates to it potentially

JackHowa (changes requested):

blocks/header-nav-chain-block/chains/header-nav-chain-block/useDebounce.js
kascote: is a leftover, I'll remove it

blocks/header-nav-chain-block/chains/header-nav-chain-block/useDebounce.js
kascote: yes sure, I don't added it as dependency because i thought we want to minimize external dependencies.

kascote:
As I put on the description I don't added tests, because the new code depends on timers and scroll.
We don't have support at this time to test with timers (I made a fix on the `update-tests-coverate` branch for this). And enzyme do not support test this kind of behavior with scroll.

JackHowa:
if the tests are in `update-tests-coverate`, can you add them to this branch?

blocks/header-nav-chain-block/chains/header-nav-chain-block/useDebounce.js
JackHowa: I'm not sure what you mean. If you need to use this package, include it as a dependency.

beltrancaliz: I think he is referring to the fact that we want to keep each block's external dependencies (non-wpmedia node modules) at a minimum, but I think in this scenario it makes sense to add `use-debounce` to `header-nav-chain-block`

kascote:
isn't a test, is a setup to add async support to tests
https://github.com/WPMedia/fusion-news-theme-blocks/pull/393/commits/9878d461433dace18320a1c5e5882172d1618f7a

JackHowa:
ok. can you add support for async tests. then add tests using that support?

JackHowa:
In summary: 

1. Remove the copy-pasted open-source code from `./useDebounce`. It is a bad practice to copy and paste large amounts of code that aren't yours.  
2. Add the package as a dependency of the block via the blocks's `package.json`. Therefore, if there are any bugfixes to that package, we can get those improvements. It also helps us if we need to see documentation for that package. Also, we're acknowledging who wrote that code so we're attributing its authors properly. 
3. Add support for async tests if you need async support to write a test via https://github.com/WPMedia/fusion-news-theme-blocks/commit/9878d461433dace18320a1c5e5882172d1618f7a
4. Write tests for this new functionality as @beltrancaliz commented

kascote:
tests added, no need to use the regenerator, `jest.fakeTimers` was enough for the tests.

beltrancaliz (approved):

PR #445, opened 2020-09-09
PEN-814 label on promo images

kascote:
[PEN-814](https://arcpublishing.atlassian.net/browse/PEN-814)

- create a function to detect the promo blocks type and implement a label to display it

# How was this tested?
- tests written

# Show Effect Of Changes

## After

- depends on this PR from engine sdk, to fix the CameraIcon
https://github.com/WPMedia/engine-theme-sdk/pull/87

### DIFF
--- a/blocks/extra-large-promo-block/features/extra-large-promo/_children/discover.js
+++ b/blocks/extra-large-promo-block/features/extra-large-promo/_children/discover.js
@@ -1,0 +1,28 @@
+
+function discoverPromoType(content) {
+  if (!content) {
+    return undefined;
+  }
+
+  let rc = '';
+
+  switch (content.type) {
+    case 'video':
+      rc = 'Video';
+      break;
+    case 'gallery':
+      rc = 'Gallery';
+      break;
+    default:
+      // eslint-disable-next-line camelcase
+      if (content.promo_items?.lead_art?.type) {
+        rc = discoverPromoType(content.promo_items.lead_art);
+      } else {
+        rc = 'other';
+      }
+  }
+
+  return rc;
+}
+
+export default discoverPromoType;
--- a/blocks/extra-large-promo-block/features/extra-large-promo/_children/discover.js
+++ b/blocks/extra-large-promo-block/features/extra-large-promo/_children/discover.js
@@ -4,25 +4,25 @@ function discoverPromoType(content) {
     return undefined;
   }
 
-  let rc = '';
+  let promoType = '';
 
   switch (content.type) {
     case 'video':
-      rc = 'Video';
+      promoType = 'Video';
       break;
     case 'gallery':
-      rc = 'Gallery';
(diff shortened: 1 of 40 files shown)


blocks/extra-large-promo-block/features/extra-large-promo/_children/discover.js
```diff
@@ -0,0 +1,28 @@
+
+function discoverPromoType(content) {
+  if (!content) {
+    return undefined;
+  }
+
+  let rc = '';
```
beltrancaliz: I'd use a more descriptive var name than `rc`

blocks/large-promo-block/features/large-promo/default.jsx
```diff
@@ -146,7 +149,7 @@ const LargePromo = ({ customFields }) => {
         <div className="row">
           {customFields.showImage
           && (
-            <div className="col-sm-12 col-md-xl-6">
+            <div className="col-sm-12 col-md-xl-6 flex-col">
```
JackHowa, line 152: was this css change deliberate?

blocks/large-promo-block/features/large-promo/_children/discover.js
```diff
@@ -0,0 +1,28 @@
+
```
beltrancaliz, line 1: `discover.js` seems to be identical across promo blocks, can we have them reuse the same file instead of having each block have its own `discover.js` and `discover.test.js`?

JackHowa:
I've noticed some changes in the chromatic snapshot that would be worth approving from a stakeholder. don't want design regressions @brandonhunty @kascote https://github.com/WPMedia/engine-theme-sdk/pull/87#pullrequestreview-485232878

RobinGiannattasio:
As part of the work in [PEN-1265](https://arcpublishing.atlassian.net/browse/PEN-1265) to standardize branches, we are deprecating the `staging` branch in favor of `canary`. So, I switched the base of this PR. Thanks!

blocks/large-promo-block/features/large-promo/default.jsx
kascote: yes, it is used on other promo blocks and on other places like
https://github.com/WPMedia/fusion-news-theme-blocks/blob/92e2562e2917f8fd3827c277eb9d4a9b3709289d/blocks/large-promo-block/features/large-promo/default.jsx#L64

on this case, the missing class was an issue with the position of the new label.

blocks/large-promo-block/features/large-promo/_children/discover.js
kascote: agree, but at this moment we don't have a place to share this common code (a block or something), after discussing the issue with @LukeMason we agree to duplicate the code by now.

beltrancaliz (approved):

PR #449, opened 2020-09-14
PEN-1309 - Syncing translations from themes-i8n repo

LukeMason:
This is an automated merge of the translations from the https://github.com/WPMedia/themes-i18n

### DIFF
--- a/blocks/byline-block/intl.json
+++ b/blocks/byline-block/intl.json
@@ -1,10 +1,12 @@
 {
-    "byline-block.by-text": {
-        "en": "By",
-        "sv": "Av"
-    },
-    "byline-block.and-text": {
-        "en": "and",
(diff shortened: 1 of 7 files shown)


kascote (approved):

kascote:
there is one more block that have an `intl.json` file, and is `header-nav-block`. That block will be deprecated some time in the future. Not sure if need to be translated too.

LukeMason:
Thanks @kascote! I added that block to the i18n repo good catch.

kascote (approved):

jgrosspietsch (approved):
Looks like the i18n repo uses 3 space tabs so it makes the diffs look like we're changing the whole file. Not something we need to care about for this PR but I'll set it to use something more conventional.

PR #450, opened 2020-09-18
PEN-1302 add focal point support

kascote:
## Description
Updated `resizer-image-block` to add the Thumbor filter for focal points when is available.

## Jira Ticket
- [PEN-1302](https://arcpublishing.atlassian.net/browse/PEN-1302)

## Acceptance Criteria
* If an image has a focal point set in the ANS data, that focal point should be respected when the image is resized by Arc Blocks
* This should apply whenever an image is resized (and the ratio is changed) – Top Table List, Results List, Numbered List, Simple List, Card List, XL/L/M/S Promo Blocks 
* The image ratio configured (or automatically set) for the image in the block should also be respected
* Tests and documentation are written to cover this functionality

## Test Steps
- Change/update an image focal point on PhotoCenter and check the new image displayed on the block.
- Pick a story and add the previous image as Featured Media (basic type) to the story, and overwrite the focal point.
- Select a gallery and change the focal point to the photo that will be the cover of the gallery. When the gallery is rendered must use the new focal point.

_Most of the photos on PhotoCenter are on 16:9 format and most of the blocks use that format, on that cases the change to the focal point is difficult to see (because there isn't a crop, only a resize). Some blocks have the option to change the image format to other values (3:2 or 4:3), this format change can generate an image crop and on that cases is more easy to spot the change to the focal point_

### Before

Screen Shot 2020-09-18 at 11 50 04

### After

Screen Shot 2020-09-18 at 11 56 14

- none

## Review Checklist


beltrancaliz (approved):
Looks good to me!

PR #475, opened 2020-10-02
PEN 1083 Storybook fix and improvement for future

JackHowa:
## Description

Basically, we need a way to mock fusion and demo the blocks. To do this, and for performance improvements, we can separate container data-fetching logic from presentational components. Storybook only needs presentational components. 

This is an incremental block to prove this works.

## Jira Ticket
https://arcpublishing.atlassian.net/secure/RapidBoard.jspa?rapidView=545&projectKey=PEN&view=planning&selectedIssue=PEN-1083&issueLimit=100&assignee=5e387d6a1b1d910e5dfd7a9b

## Acceptance Criteria
The mocking of fusion functionality in storybook is somewhat hack-ish. In essence, we’re still investigating a better solution. It’s mostly been put on hold. If we want, in this ticket, I can investigate it more. In essence, we have to change some existing functionality to “fake” fusion data. This is not a best practice. 

## Test Steps
- Run `npm run storybook`
- Play with headline text input and fonts

### Before
- Had to hack storybook to use it

### After
- Not hacky and better dev experience for storybook

- 

## Review Checklist
- [ x] Confirmed all the test steps above are working
- [x ] Confirmed there are no linter errors
- [x ] Confirmed this PR has reasonable code coverage
  - [x ] Confirmed this PR has unit test files
  - [ x] Ran `npm test`, made sure all tests are passing
  - [x ] If the amount of work to write unit tests for this change are excessive,
- [x ] Confirmed relevant documentation has been updated/added.


beltrancaliz:
@JackHowa It works well when I tested it locally. However is this the right path forward? Is it worth changing all blocks to adapt them to storybook even though it might introduce some regressions? With more complex blocks I can see removing fusion mocks as somewhat problematic but this is definitely a great starting point.

JackHowa:
Thanks @beltrancaliz.

JackHowa:
> However is this the right path forward? Is it worth changing all blocks to adapt them to storybook even though it might introduce some regressions?

I think these are good questions. With good testing, we should not see regressions. And, with storybook setup in the future, we'll see fewer regressions moving forward. For example, storybook includes news theme css crucially. Right now, it feels impossible to test them. 

> With more complex blocks I can see removing fusion mocks as somewhat problematic but this is definitely a great starting point.

I kind of agree with you that it might be dangerous moving forward with more complex blocks. However, I think if we start small, we can work our way up. 

Also, storybook is a decision that product wants for client documentation. And, without this pattern, storybook is impossible and more dangerous (with how it's currently injected). 

On a totally separate side note, reworking on these could also have massive performance benefits. Rerendering with complex props and object destructuring is extremely expensive. I think this could be a massive opportunity. 

^ this is a bit preachy, but helpful at least for me to think about. we can discuss this moving forward and I can present. Joe's onboard with this pattern as well, which is good. the other pattern of injecting props was, in hindsight, a bit of a hack

blocks/headline-block/features/headline/default.jsx
```diff
@@ -8,16 +9,42 @@ const HeadlineHeader = styled.h1`
...
-    !!(content && content.headlines && content.headlines.basic) && (
-      <HeadlineHeader className="headline" dangerouslySetInnerHTML={{ __html: content.headlines.basic }} primaryFont={getThemeStyle(arcSite)['primary-font-family']} />
-    )
-  );
+// container handles data fetching, connection to fusion
+const HeadlineContainer = () => {
+  // get headlines basic
+  const { globalContent } = useFusionContext();
+
+  const { basic: headlineString = '' } = globalContent?.headlines || {};
```
RobinGiannattasio, line 32: Definitely can see some helpers happening here eventually! I think as we standardize things in the container method, we'll see a real pattern emerge around how we handle the resulting data (since when we call it will now be happening consistently!)

blocks/headline-block/features/headline/default.test.jsx
```diff
@@ -27,14 +27,16 @@ describe('the headline feature for the default output type', () => {
...
 
       expect(wrapper.find('h1')).toHaveClassName('headline');
+      // checking for styled component class
       expect(wrapper.find('h1').hasClass(/sc-/)).toBe(true);
     });
 
     it('should dangerously set the innerHTML to the headline content', () => {
       const { default: Headline } = require('./default');
-      const wrapper = shallow(<Headline />);
+      const wrapper = mount(<Headline />);
```
RobinGiannattasio, line 36: I occasionally worry about using mount because it tends it falsely increase coverage when running tests. Meaning, it'll mount ALL the things, and even though things aren't asserted, it still has run the code and considers itself covered.

I prefer to use shallow and if I *need* to mount a child node, I try and use the `dive` method.

https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/dive.html

I don't think you have to change this, just worth having the conversation around testing methods :D

blocks/headline-block/index.story.jsx
```diff
@@ -1,39 +1,18 @@
...
-      },
-    },
-  });
+  const headlineString = text('headlineString', 'Man Bites Dog');
+  // if we want to allow custom fonts https://stackoverflow.com/a/63128475/7491536
+  const primaryFont = text('primaryFont (web-safe)', 'Arial');
 
   return (
-    <Headline useInjectedFusionContext={newData} />
+    <Headline headlineString={headlineString} primaryFont={primaryFont} />
```
RobinGiannattasio, line 16: So much cleaner!

clean dino

blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import PropTypes from '@arc-fusion/prop-types';
```
RobinGiannattasio, line 2: 👍

RobinGiannattasio (approved):
I think this pattern is great! Also, thinking about performance from last sprint that I'll talk about on Thursday, I think it would be good for us to evaluate each block and see if we are doing the following:

- Using minimal HTML and removing unnecessary divs/nodes that we are using for styling
- Make sure we are not rendering empty nodes. So, here, for example. We are not rendering anything if there is not a `headlingString`, which is correct.
- See where we can reduce react re-renders per component.

I think we can create some sensible peformance/storybook improvement tickets for groups of blocks and attack them together!

VERY EXCITED.

celebrate

blocks/headline-block/features/headline/default.test.jsx
JackHowa: huh, haven't seen this discussion. Yeah good to talk about

blocks/headline-block/index.story.jsx
JackHowa: totally agree -- thanks

blocks/share-bar-block/features/share-bar/default.jsx
JackHowa: no worries -- was breaking storybook

PR #502, opened 2020-10-15
Update blocks dependencies

kascote:
updated the dependencies on all story-feed blocks.
* most of the cases was having depency against `engine-sdk` when is not needed.
* added dependency on `resizer-image-block`
* remove dependency on `content-source_story-feed_sections-v4` from `story-feed-query`

tested each cs locally in PB content debugger


JackHowa (approved):
good work Nelson

PR #506, opened 2020-10-16
Perf fix use peer deps ignore test files

JackHowa:
Problem: 
- was seeing in compiler logs and internal service error because device ran out of space installing dependencies 

Solution:

- Use peerDeps rather than deps for already-installed packages and reused ones (engine theme sdk, news theme css, other blocks, styled components)
- Kept dependencies actually needed (use-debounce, react-oembed --> although this is in news-theme feature pack)
- Ignored some test files and big mocks (especially resizer block)
 
In future: 

- [ ] Document using peerDependencies. When to use, when not to. Can we make further improvements in blocks installer?
- [ ] Create .npmignore in each block. I tried with  ` find . -0 -type d -exec touch {}/.npmginore \;\n` (got pretty close)
-> make each block ignore *.test.js and *.test.jsx and any other extraneous deps 

Test:
- See in corecomponents along with https://github.com/WPMedia/Fusion-News-Theme/pull/128 bundle 
-> no errors in compiler either and working in corecomponents 
- Run `npx fusion start -l` to test locally (working as epected for me)

Screen Shot 2020-10-16 at 6 07 00 PM


beltrancaliz (approved):
Tested locally and looks good to me. Great job!

PR #523, opened 2020-10-28
PEN-1358 - update small promo blocks to display 3 and 4 items per row

kascote:
## Description
Update small promo blocks on `top-table-list-block` to support 3 and 4 items per row, and update the layouts for they.

## Jira Ticket
- [PEN-1358](https://arcpublishing.atlassian.net/browse/PEN-1358)

## Acceptance Criteria
PageBuilder users can configure Small items in the Top Table List to display 3 or 4 stories per row by using the Stories per row custom field (in addition to 1 or 2 stories per row, which is already supported)

If 3 stories per row or 4 stories per row is selected, this should only display on the Desktop/Tablet breakpoints – on the mobile breakpoint, it should display 2 per row

## Test Steps
- add a `top-table-list-block` to a page
- setup it to show for example 10 small promo blocks
- update the custom fields to show different numbers of items per row
- verify that the layout change
- later setup the small promo blocks to show 4 items per row, and change browser viewport to mobile, and verify that the layout change to 1 item per row.

### use case where the images are setup to 4:3 ratio and the fallback image is on 16:9 
a followup ticket will address this issue (send local images throw thumbor)

- none

## Review Checklist

### DIFF
--- a/blocks/top-table-list-block/features/top-table-list/_children/item-title-with-right-image.test.jsx
+++ b/blocks/top-table-list-block/features/top-table-list/_children/item-title-with-right-image.test.jsx
@@ -20,34 +20,10 @@ const config = {
   showBylineMD: true,
   showDateMD: true,
   showHeadlineSM: true,
-  // headlinePositionSM: 'above',
   showImageSM: true,
 };
 
-// const headBelowConfig = {
-//   showOverlineXL: true,
-//   showHeadlineXL: true,
-//   showImageXL: true,
-//   showDescriptionXL: true,
-//   showBylineXL: true,
-//   showDateXL: true,
-//   showOverlineLG: true,
-//   showHeadlineLG: true,
-//   showImageLG: true,
-//   showDescriptionLG: true,
-//   showBylineLG: true,
-//   showDateLG: true,
-//   showHeadlineMD: true,
-//   showImageMD: true,
-//   showDescriptionMD: true,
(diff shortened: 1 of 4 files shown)


blocks/top-table-list-block/features/top-table-list/_children/item-title-with-right-image.test.jsx
```diff
@@ -84,153 +59,236 @@ describe('item title with right image block', () => {
...
+    const id = 'test';
+    const { default: ItemTitleWithRightImage } = require('./item-title-with-right-image');
+
+    const wrapper = mount(
+      <ItemTitleWithRightImage
+        imageURL={imageURL}
+        itemTitle={itemTitle}
+        primaryFont={primaryFont}
+        id={id}
+        /* eslint-disable-next-line */
```
beltrancaliz, line 235: we might want to apply this to the whole file if we have it in a few places

beltrancaliz (approved):

blocks/top-table-list-block/features/top-table-list/_children/item-title-with-right-image.test.jsx
kascote: but that will disable the eslint on the whole file, not sure we want that.
The problem here is that eslint complaint about Object.assign and suggest to use the deconstructor way, but other nasty error happens using the deconstructor.

PR #527, opened 2020-10-29
Pen 1304 404 support tags content source

kascote:
## Description
throw 404 error on content source when tag not found

## Jira Ticket
- [PEN-1304](https://arcpublishing.atlassian.net/browse/PEN-1304)

## Acceptance Criteria
If a Tags page is requested with a tag slug that does not exist, the page redirects to a 404 Not Found page

## Test Steps
- go to `localhost/tag/food` and verify that renders correctly
- then go to `localhost/tag/nonsense` and verify that the content source throw 404 error

### After
Screen Shot 2020-10-29 at 11 41 29

- none

## Review Checklist

### DIFF
--- a/blocks/tags-content-source-block/sources/tags-api.js
+++ b/blocks/tags-content-source-block/sources/tags-api.js
@@ -1,8 +1,26 @@
+const make404 = () => {
+  const error = new Error('Not found');
+  error.statusCode = 404;
+  throw error;
+};
+
 export default {
   resolve({ slug }) {
     return `/tags/v2/slugs?slugs=${slug}`;
   },
   params: {
     slug: 'text',
+  },
+  transform: (data) => {
+    if (!data || !data.Payload) {
+      make404();
+    }
+
+    const hasValues = data.Payload.some((ele) => !!ele);
+    if (hasValues) {
+      return data;
+    }
+
+    return make404();
   },
 };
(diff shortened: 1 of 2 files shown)


blocks/tags-content-source-block/sources/tags-api.js
```diff
@@ -1,8 +1,26 @@
...
 export default {
   resolve({ slug }) {
     return `/tags/v2/slugs?slugs=${slug}`;
   },
   params: {
     slug: 'text',
   },
+  transform: (data) => {
+    if (!data || !data.Payload) {
+      make404();
```
JackHowa, line 16: would you want to return here as well?

JackHowa:
```suggestion
      return make404();
```

blocks/tags-content-source-block/sources/tags-api.test.js
```diff
@@ -23,3 +23,54 @@ describe('the tags content source block', () => {
     });
   });
 });
+
+describe('the transform', () => {
+  const error = new Error('Not found');
+  error.statusCode = 404;
+
+  it('must return the original data if something found', () => {
```
JackHowa, line 31: good work with tests

blocks/tags-content-source-block/sources/tags-api.js
kascote: the return at the end is only to comply with eslint, is not needed because the function never returns, throw an error

JackHowa: right right

JackHowa (approved):
Good work Nelson

PR #529, opened 2020-10-30
Hotfix peer deps blocks to beta 

JackHowa:
https://arcpublishing.atlassian.net/secure/RapidBoard.jspa?rapidView=545&projectKey=PEN&modal=detail&selectedIssue=PEN-1520&assignee=5e387d6a1b1d910e5dfd7a9b

Use peerDependencies when we can rather than dependencies. As a note, this is already working in canary ... see rca https://github.com/WPMedia/fusion-news-theme-blocks/wiki/Bundle-Size-Internal-Service-Error-On-Corecomponents

prev explanation (I removed the non-peer deps experiment):

> - Similarly instrumenting peerDeps to https://github.com/WPMedia/fusion-news-theme-blocks/wiki/Bundle-Size-Internal-Service-Error-On-Corecomponents
> - except, we have designated deps (not peers) for showing that we can install a dependency version of engine theme sdk
> - I also removed content sources like I did previously for canary as they are installed by the fusion-news-theme and working as expected in corecomponents
> - Designated the correct news-theme-css version. But just because we will be installing the correct version at the top level of beta and want to make the peer dependencies match and not throw a warning
> - For Monday

### DIFF
--- a/blocks/alert-bar-block/package.json
+++ b/blocks/alert-bar-block/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "cookie": "^0.4.1",
-    "@wpmedia/engine-theme-sdk": "stable"
+    "@wpmedia/engine-theme-sdk": "beta"
   },
   "gitHead": "8afcaeae3d9a0cc126a4a22a9b7fe766669aa88e"
 }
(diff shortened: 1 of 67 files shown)


JackHowa:
working as expected https://corecomponents.arcpublishing.com/pf/all-blocks/?_website=washpost&d=532

seans887 (approved):
LGTM, I saw Jack test these changes out in corecomponents and they seem to work. Would like to get 1 more technical approval if possible.

jgrosspietsch (approved):
:cool:

This definitely is what it says it is.

PR #542, opened 2020-11-05
ci: Add test blocks feature on pr

JackHowa:
## Description
Test coverage and linting on pr 

## Jira Ticket
- [PEN-1339](https://arcpublishing.atlassian.net/secure/RapidBoard.jspa?rapidView=545&projectKey=PEN&modal=detail&selectedIssue=PEN-1370&assignee=5e387d6a1b1d910e5dfd7a9b-)

## Acceptance Criteria
Run tests with coverage https://github.com/WPMedia/fusion-news-theme-blocks/blob/stable/package.json#L22

there are already github actions that you can base your implementation off of https://github.com/WPMedia/fusion-news-theme-blocks/actions?query=workflow%3A"Canary+block"

the tests should also be seen in the prs against these branches as well as a check

if all the tests do not pass, the pr should not be able to be merged 

additionally, if the tests do not pass, the publish action should not happen on push to canary and stable branches

Make sure you run npm i not `npx lerna bootstrap -y` because the tests rely on top-level node modules. You will get a rule of hooks error if there are nested dependencies

You may need to use secrets in github ui to run the github actions for npm i

make sure the minimum coverage check command is run as well https://github.com/WPMedia/fusion-news-theme-blocks/blob/stable/package.json#L22

the packages should not be published if the coverage check doesn’t pass

the pr should not be able to merged without the coverage check

## Test Steps
- Make sure tests pass on this branch (meta, right?)

### DIFF
--- a/.github/workflows/test-coverage-blocks.yml
+++ b/.github/workflows/test-coverage-blocks.yml
@@ -28,13 +28,13 @@ jobs:
 
       - name: Install dependencies if lockfile changed
         if: steps.cache.outputs.cache-hit != 'true'
-        run: npm install
+        run: npm ci
         env:
           NODE_AUTH_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
 
       # Remove nested modules to prevent multiple react versions
       - name: Run tests with coverage
-        run: npx lerna clean -y && npm run test:coverage
+        run: npm run test:coverage
 
       - name: Check linting
         run: npm run lint
--- a/.github/workflows/test-coverage-blocks.yml
+++ b/.github/workflows/test-coverage-blocks.yml
@@ -32,7 +32,6 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
 
-      # Remove nested modules to prevent multiple react versions
       - name: Run tests with coverage
         run: npm run test:coverage
 
--- a/.github/workflows/test-coverage-blocks.yml
+++ b/.github/workflows/test-coverage-blocks.yml
@@ -19,15 +19,15 @@ jobs:
           node-version: "12"
           registry-url: "https://npm.pkg.github.com"
 
-      - name: Cache NPM files
+      - name: Cache Node Modules
+        id: cache
         uses: actions/cache@v2
         with:
-          path: ~/.npm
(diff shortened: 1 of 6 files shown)


JackHowa:
@kascote hopefully this is helpful for ya -- glad the warnings will be reduced 🤞

.github/workflows/test-coverage-blocks.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Cache Node Modules
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package-lock.json') }}
+
+      - name: Install dependencies if lockfile changed
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: npm install
```
brentmiller1973: Wonder since this is a Lerna repo if we should do `npm ci` instead of just a basic install?  `npm ci` is a clean and then install.

brentmiller1973 (commented):
Nice Jack...only suggestion is I think we should use npm ci instead of npm install.  You know how when working locally with lerna we have to sometimes remove the node_modules directory and then re-install?  I think npm ci would replicate this process so we don't have erroneous test results

.github/workflows/test-coverage-blocks.yml
JackHowa: huh not sure, can try it out. don't really see a difference in description between the two https://docs.npmjs.com/cli/v6/commands/npm-ci#description

JackHowa: interesting yeah, updated

JackHowa:
with `npm ci` and with a fresh install each time, as in canary, that the caching is actually doing anything: 

```
      - name: Cache NPM files
        uses: actions/cache@v2
        with:
          path: ~/.npm
          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
          restore-keys: |
            ${{ runner.os }}-node-

brentmiller1973 (approved):

.github/workflows/test-coverage-blocks.yml
JackHowa: thanks @visionbegin -- working

PR #543, opened 2020-11-06
removed warnings from default-output-block tests

kascote:
## Description
- removed warnings from `default-output-block` tests
- created `uniqid` function to generate the keys for the scripts inserted

### Before
lots, and lots of 

### After
none!

- none

## Review Checklist


blocks/default-output-block/output-types/default.jsx
```diff
@@ -3,13 +3,14 @@ import { playerRoot, videoOrg } from 'fusion:environment';
...
 import { MetaData } from '@wpmedia/engine-theme-sdk';
+import uniqid from './uniqid';
 import './default.scss';
 
 const powaBoot = `${playerRoot}/prod/powaBoot.js?=org=${videoOrg}`;
 const powaDrive = `${playerRoot}/prod/powaDrive.js?org=${videoOrg}`;
 
 const injectStringScriptArray = (scriptStringArray) => scriptStringArray.map((scriptString) => (
-  <script dangerouslySetInnerHTML={{ __html: scriptString }} />
+  <script key={uniqid('script')} dangerouslySetInnerHTML={{ __html: scriptString }} />
```
JackHowa: can we find some other way of a key for this? like the key of the env `gaScript` or something like that. We shouldn't need a function like uniqid

blocks/default-output-block/output-types/__tests__/default.test.jsx
```diff
@@ -272,7 +327,10 @@ describe('chartbeat render conditions', () => {
     }))));
 
     const { default: DefaultOutputType } = require('../default');
-    const wrapper = shallow(<DefaultOutputType deployment={jest.fn()} metaValue={jest.fn().mockReturnValue('article')} />);
+    const wrapper = shallow(
+      // eslint-disable-next-line react/jsx-props-no-spreading
```
JackHowa: I believe we have an ignore for this rule globally in the eslint config. if not, can you add one. we use props spreading all over the place and aren't against this as a team

blocks/default-output-block/output-types/__tests__/default.test.jsx
```diff
@@ -7,6 +7,13 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+const dummyComp = () => <meta content="dummy" />;
+const mockFuntions = {
+  Libs: dummyComp,
```
JackHowa, line 12: Not sure I totally understand this mock. can you elaborate?

blocks/default-output-block/output-types/default.jsx
kascote: I think is not so easy. That scripts came from an array on `blocks.json`. They are custom script the site can add.

https://github.com/WPMedia/Fusion-News-Theme/blob/master/blocks.json#L136
https://github.com/WPMedia/fusion-news-theme-blocks/blob/c7da8bccfb8ee711f88274fa1d4f937b42c51340/blocks/default-output-block/output-types/default.jsx#L90

I was thinking on use some kind of hash over the script content but I think overkill for this.

blocks/default-output-block/output-types/__tests__/default.test.jsx
kascote: I'm mocking the components that default-output-block use to render the libs, css and fusion data.
https://github.com/WPMedia/fusion-news-theme-blocks/blob/c7da8bccfb8ee711f88274fa1d4f937b42c51340/blocks/default-output-block/output-types/default.jsx#L182

there was warnings on the tests because those components wasn't received.

blocks/default-output-block/output-types/default.jsx
JackHowa: ah, I see now. huh, not sure

JackHowa:
Script tags shouldn't be re-rendering or really be changing. If you want, we can ignore this or quiet it. The reason for keys don't really apply here

blocks/default-output-block/output-types/default.jsx
JackHowa: if you remove the key and uniqid script update, I think this pr will be able to remove lots of the warnings based on your before screenshot

kascote:
the warning on missing keys can't be disabled (1). One solution could be spreading the properties over the `React.createElement` function (2) instead of use JSX.
Other solution could be left the current implementation or simplify it using the array index, just to comply with the key requirement. I agree this is odd but I think is better than has the warning sitting around. 
May be I'm wrong, but output-type is always SSR and on this case the keys are not meaningfull. I think the uniqId or the array index are ok. 

(1) https://github.com/facebook/react/issues/12567
(2) https://github.com/facebook/react/issues/12567#issuecomment-414061454

JackHowa:
yeah index should be ok functionally. were there warnings for this array dangerously set array? Looks like the screenshot was related to something else

kascote:
yeah :( the screenshot only show the warning related to the missing components (the mock I did).
this is the one related to the keys:

JackHowa:
in my mind, this is preferable 

```
// import uniqid from './uniqid';
import './default.scss';

const powaBoot = `${playerRoot}/prod/powaBoot.js?=org=${videoOrg}`;
const powaDrive = `${playerRoot}/prod/powaDrive.js?org=${videoOrg}`;

const injectStringScriptArray = (scriptStringArray) => scriptStringArray.map(
  (scriptString, index) => (
    // no good way of getting keys for this
    // index used to remove warnings
    // this key will not affect performance or issues with changing order
    // eslint-disable-next-line react/no-array-index-key
    <script key={index} dangerouslySetInnerHTML={{ __html: scriptString }} />
  ),
);

**Comment by JackHowa** [comment]
^ no test warnings: 

```
howardj@it-20750 fusion-news-theme-blocks % npm run test -- -- blocks/default-output-block/output-types/__tests__/default.test.jsx

> fusion-news-theme-blocks@ test /Users/howardj/sites/fusion-news-theme-blocks
> jest "--" "blocks/default-output-block/output-types/__tests__/default.test.jsx"

 PASS  blocks/default-output-block/output-types/__tests__/default.test.jsx
  the default output type
    ✓ should render (860ms)
  renders a page
    ✓ should have a head (76ms)
    ✓ should have a body (73ms)
    ✓ should have script tags (63ms)
    ✓ should have link tags (53ms)
    ✓ should have a MedataData component (62ms)
    ✓ MedataData should receive twitterUsername (122ms)
  root html layout
    ✓ html must have only head and body tags (66ms)
  head content
    ✓ must render Google Tag Manager script (63ms)
    ✓ must render Google Analytics script (52ms)
    ✓ must render custom script (59ms)
    ✓ must render custom font url (59ms)
    ✓ must not render nested scripts (64ms)
  head content without properties
    ✓ must not render chartbeat code (60ms)
  chartbeat render conditions
    ✓ must not render chartbeat code when properties are missing (60ms)
    ✓ must not render chartbeat code when chartBeatAccountId property is missing (65ms)
    ✓ must not render chartbeat code when chartBeatDomain property is missing (61ms)
    ✓ must not render chartbeat code when both properties are empty (77ms)
    ✓ must render chartbeat code when chartBeatDomain and chartBeatAccountID properties are present (58ms)
  comscore render conditions
    ✓ must not render comscore code when property is missing (59ms)
    ✓ must render comscore code when comscoreID site property is present (59ms)

Test Suites: 1 passed, 1 total
Tests:       21 passed, 21 total
Snapshots:   0 total
Time:        3.598s, estimated 4s
Ran all test suites matching /blocks\/default-output-block\/output-types\/__tests__\/default.test.jsx/i.
howardj@it-20750 fusion-news-theme-blocks %

JackHowa:
wait, we are just rendering the arr... I would expect to spread that array to get the react elements: 

```
        {...injectStringScriptArray(dangerouslyInjectJS)}

kascote:
React will render the elements, no the array, the spreading is implicit
https://reactjs.org/docs/lists-and-keys.html#rendering-multiple-components

JackHowa:
right right, was questioning the script tag in the head tag for a second

> React will render the elements, no the array, the spreading is implicit
https://reactjs.org/docs/lists-and-keys.html#rendering-multiple-components

JackHowa (approved):
sorry this was a hassle, thank you

PR #546, opened 2020-11-06
Pen 1459 video promo live

amyerz:
## Description
- update the live customfields from checkbox to text input
- if input box is empty, live label should not be shown

## Jira Ticket
- [PEN-1459](https://arcpublishing.atlassian.net/browse/PEN-1459)

## Acceptance Criteria
On the Video Promo Block, remove the existing Live boolean Custom Field, and instead implement a string (text) custom field called Alert badge 

If a value is entered in this field, that exact text should appear above the video (aligned left), using the label specs in the Zeplin designs (using the Danger red color, which is also used for the Alert Bar) 

Note – the text should render exactly as it is input into the Custom Field – don’t automatically capitalize it or anything like that

If no value is entered in this field, then the Alert badge should not appear at all

## Test Steps
1. on the theme repo, update .env file:
- [ ] link your local engine-theme-sdk, e.g. ENGINE_SDK_REPO={PATH/TO/LOCAL/engine-theme-sdk}
- [ ] add videoOrg=prod
- [ ] add videoOrg=corecomponents

2. under block repo
create a new page and add video promo block

3. test the `Live` customfield, if no value is entered, the label should not be shown

4. entered a sample value to the `Live` field, the text should match on the label without any capitalization etc
Screen Shot 2020-11-06 at 11 00 21 AM

### Before
- `Live` customfield is a checkbox:
Screen Shot 2020-11-06 at 2 29 02 PM

### After
- `Live` customfield updated to text input box
Screen Shot 2020-11-06 at 11 00 31 AM

_Examples of dependencies or side effects are:_

## Review Checklist
- [ ] Confirmed all the test steps above are working
- [ ] Confirmed there are no linter errors
- [ ] Confirmed this PR has reasonable code coverage
  - [ ] Confirmed this PR has unit test files
  - [ ] Ran `npm test`, made sure all tests are passing

### DIFF
--- a/blocks/video-promo-block/README.md
+++ b/blocks/video-promo-block/README.md
@@ -20,7 +20,7 @@ PageBuilder users can configure:
 | **ratio** | no | number | the aspect ratio of the video used by the powa player  |
 | **title** | no | string | the title of the component |
 | **description** | no | string | the description of the component  |
-| **live** | no | boolean | if set to true, a `LIVE VIDEO` label will show up on top of the component |
+| **live** | no | string | a `LIVE VIDEO` label will show up on top of the component if value is not empty |
 
 ## ANS Schema
 ANS of the video type
--- a/blocks/video-promo-block/README.md
+++ b/blocks/video-promo-block/README.md
@@ -10,6 +10,7 @@ PageBuilder users can configure:
 - Description
 - LIVE
 - Autoplay
+- Playthrough
 
 ## Props
 | **Prop** | **Required** | **Type** | **Description** |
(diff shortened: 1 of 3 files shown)


JackHowa (changes requested):
Can you write a test in case live is the value `true`. I imagine that when this is deployed some preexisting uses of this will have that value. I'd be worried that `{live && <p>{live}</p>}` would render `<p>true</p>`, trying to stringify a previous block. we want our blocks to be able to handle previous versions of the props

PR #549, opened 2020-11-06
removed article-tag tests warnings 

kascote:
## Description
removed article-tag tests warnings

- none

## Review Checklist


JackHowa (changes requested):
Good thought, but the key should be unique to the content, not generated. please see https://reactjs.org/docs/reconciliation.html#keys . you had a good example when you found the _id property or some other unique key

kascote:
I was reviewing a couple stories and no one has `_id` as key (but the ANS spec say must have one). And the slug can be empty too (!?) according to the codebase. 
do you think a string hash over `tag.text` is better ? string hash over all the tag fields ? (slug, name, description)

JackHowa:
we want the tests to mimic the actual data as much as possible. so, if the ans spec the examples must have an `_id` property, you can add those to the examples to test for. If you update the examples to include the _id to be unique, that should be the best solution

sorry this is taking time; your hashing of slug, etc is definitely clever, but I don't think it's the best solution to this.

kascote:
the tests mimic the actual data I saw on a lot stories I retrieved to check the actual values, and no one has the `_id` field. For that reason I'm not using it on the tests or to calculate the hash.
if for some sites the `_id` could exists and for others no, I think normalizing on use a hash is a fine solution. Or I can test if `_id` exists and use it instead of calculate the hash. (?)

JackHowa:
tag.text is unique right? could we use that alone? not sure if the hash is necessary. or the href -- both of those fields seem like they have to exist and are unique. I'd recommend: 

```jsx
          return <Tags key={tag.text} className="tags" href={href} primaryColor={primaryColor || defaultBackgroundColor} primaryFont={primaryFont}>{tag.text}</Tags>;
```

JackHowa:
no test id issues with this tagging with text: 

```
howardj@it-20750 fusion-news-theme-blocks % npm run test -- blocks/article-tag-block/features/tag/default.test.jsx

> fusion-news-theme-blocks@ test /Users/howardj/sites/fusion-news-theme-blocks
> jest "blocks/article-tag-block/features/tag/default.test.jsx"

 PASS  blocks/article-tag-block/features/tag/default.test.jsx
  the article tag block
    when the global content has an array of tags in its taxonomy
      ✓ should render a parent container for the tags (75ms)
      ✓ should render a tag element for each tag in the array (11ms)
      ✓ should render tags with their correct href (8ms)
    when the global content has a property called taxonomy which contains tags array with out slugs
      ✓ should render tags with correct href (11ms)
    when the global content has an empty tags array
      ✓ should not render anything (6ms)
    when the global content does not have a taxonomy property
      ✓ should not render anything (6ms)

Test Suites: 1 passed, 1 total
Tests:       6 passed, 6 total
Snapshots:   0 total
Time:        2.071s, estimated 6s
Ran all test suites matching /blocks\/article-tag-block\/features\/tag\/default.test.jsx/i.
howardj@it-20750 fusion-news-theme-blocks %

JackHowa (approved):
no warnings with tag text

JackHowa:
Thanks for bringing this up @kascote !

PR #559, opened 2020-11-16
Add back dependency not assumed from feature pack

JackHowa:
- content source removed but should not have been assumed to be in feature pack deps


jgrosspietsch (approved):
:cool:

PR #564, opened 2020-11-17
Compression option (canary)

jgrosspietsch:
See https://github.com/WPMedia/fusion-news-theme-blocks/pull/557


JackHowa:
No longer necessary for canary blocks fix, can backport beta fix when it's validated with rc

PR #629, opened 2020-12-16
Top-level dependencies install

JackHowa:
- Not sure totally what's amiss, but it seems like we need to install dependencies at the top-level for any dep relied upon in the tests 
- Had issues with current canary branch, running rm -rf node_modules && npm i && npx lerna clean -y && npm t
- after these updates I didn't have those issues
- My hunch is that rogue package-lock files are not being updated and lingering to make sure tests remain working. Because `npm ci && npm t` still works as expected. `npm ci` installs exactly what's in the package-lock


JackHowa:
seems like Brent was having issues too

JackHowa:
to test `npx lerna clean -y && rm -rf node_modules && npm i && npx lerna clean -y && npm t`

beltrancaliz (approved):

PR #631, opened 2020-12-16
1.5 release

JackHowa:
- publish latest 
- then github action will tag that as stable 
https://github.com/WPMedia/fusion-news-theme-blocks/runs/1475429902?check_suite_focus=true
- canary publish is a false negative check


JackHowa (commented):
looks good

PR #637, opened 2020-12-18
Perf remove mock feature files

JackHowa:
- remove mocks files from publish within chains, sources, and features
- again, used automated file creation and handy vs code


JackHowa (commented):
good job Jack

PR #642, opened 2020-12-24
PEN-1620 Update Queryly integration

abwizeline:
We built an integration with Queryly (a search vendor) in PEN-1488: [LIB] Queryly search integration READY FOR DEPLOYMENT. In this ticket, we just need to update one part of that code based on feedback from Queryly. 

Task
In the code implemented in PEN-1488: [LIB] Queryly search integration READY FOR DEPLOYMENT, instead of this:

 document.querySelectorAll('main, .header-fix-nav')

use this: 

document.querySelectorAll("#fusion-app")
## Description
_Information about what you changed for this PR_

## Jira Ticket
- [PEN-1620](https://arcpublishing.atlassian.net/browse/PEN-1620)

## Acceptance Criteria
document.querySelectorAll("#fusion-app")

### Before
document.querySelectorAll('main, .header-fix-nav')

### After
document.querySelectorAll("#fusion-app")

_Examples of dependencies or side effects are:_

## Review Checklist


blocks/default-output-block/output-types/default.jsx
```diff
@@ -46,7 +46,7 @@ const chartBeatCode = (accountId, domain) => {
 
 const querylyCode = (querylyId, querylyOrg, pageType) => {
   const querylyInit = `
-    queryly.init("${querylyId}", document.querySelectorAll("main, .header-fix-nav"));
+    queryly.init("${querylyId}", document.querySelectorAll("#fusion-app"));
```
beltrancaliz, line 49: This will add queryly to every page with the default output type even if the client doesn't use the queryly integration. Is that the desired behavior?

scarothers: @beltrancaliz I don't see that happening – I only see the queryly script on websites that have **querylyId** set in blocks.json. 

Example – **querylyId** is set on The Gazette: 
https://github.com/WPMedia/Fusion-News-Theme/blob/master/blocks.json#L265
And I see the queryly script when I look in the Network tab on that site: 
https://corecomponents.arcpublishing.com/pf/all-blocks/?_website=the-gazette
Screen Shot 2021-01-04 at 1 49 00 PM


But **querylyId** is _not_ set on The Prophet: 
https://github.com/WPMedia/Fusion-News-Theme/blob/master/blocks.json#L200
And I do not see the script loading on that page: 
https://corecomponents.arcpublishing.com/pf/all-blocks/?_website=the-prophet
Screen Shot 2021-01-04 at 1 49 07 PM

This seems to be working as expected. The update in this PR is just to update the selector that's used (per some feedback from Queryly). I don't think that should change how the script is loaded, right?

beltrancaliz: @scarothers got it, I'm assuming the selector update it will make it behave the same way

beltrancaliz (approved):

PR #716, opened 2021-02-02
Css variable global sass

JackHowa:
- css variables and using scss tokens 
- todo: does this work with multi-site? @jgrosspietsch @matthewroach


JackHowa:
todo: look into alternative

PR #731, opened 2021-02-09
Revert default output removal

JackHowa:
- these styles actually used (not sure how) 
- see where margin is missed https://corecomponents.arcpublishing.com/pf/all-blocks/?_website=washpost&d=657


JackHowa:
font-weight normal;

actually used to remove margin

matthewroach (approved):

PR #793, opened 2021-03-29
Masthead styled components version change

matthewroach:
Updating version of styled-component package listed in peerDependencies.

1. It's higher than all other blocks
2. Causes some local linking issues


JackHowa (approved):
good find

matthewroach:
Thanks to @badmintonkid for finding when fixing some Fusion block issues :)

PR #805, opened 2021-04-01
Revert TMEDIA-108

matthewroach:
Reverting TMEDIA-108


blocks/shared-styles/index.story.jsx
```diff
@@ -29,7 +29,7 @@ export const primaryFont = () => {
   );
 };
 
-export const secondaryFont = () => {
+export const secondaryFontBlock = () => {
```
JackHowa, line 32: just saw this, rename

JackHowa: not sure if this is intentional?

JackHowa (commented):
we should note this revert bc there's a lot of good stuff in here

JackHowa (approved):

PR #816, opened 2021-04-08
TMEDIA-2 small promo code duplication

matthewroach:
## Description

Make reusable promo headline and promo image components in shared styles to allow better reuse between small promo and small manual promo blocks. Also allow for other promo's to move over to consume them to reduce duplication of code between promo blocks.

Each of the reusable components can consume ANS content or custom content to allow them to work for both promo and manual promo blocks

In updating the code there are some slight visual differences with padding and gutters - will the acceptance criteria states no visual differences. The differences are fixing issues with the existing implementation of adding too much padding.

## Jira Ticket
- [TMEDIA-2](https://arcpublishing.atlassian.net/browse/TMEDIA-2)

## Acceptance Criteria
* Small Promo Block and Small Manual Promo Block should have no visible changes or visual regressions for the user
* Small Promo Block and Small Manual Promo Block should use a shared set of components to output promo content
* Small Promo Block custom fields remain the same via PageBuilder 
* Small Manual Promo Block custom fields remain the same via PageBuilder

## Test Steps

1. Checkout this branch `git checkout TMEDIA-2-small-promo-code-duplication`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/small-promo-block,@wpmedia/small-manual-promo-block,@wpmedia/shared-styles,@wpmedia/double-chain-block`
3. Add small promo and small manual promo blocks to a page and verify all still works as expected
4. Validate existing pages using small promo's still work as expected - http://localhost/pf/test-all-promos/?_website=the-gazette

### Before

TMEDIA-2-before

### After

TMEDIA-2-after

## Author Checklist
_The author of the PR should fill out the following sections to ensure this PR is ready for review._
- [X] Confirmed all the test steps a reviewer will follow above are working.
- [X] Confirmed there are no linter errors. Please run `npm run lint` to check for errors. Often, `npm run lint:fix` will fix those errors and warnings.
- [X] Ran this code locally and checked that there are not any unintended side effects. For example, that a CSS selector is scoped only to a particular block.
- [X] Confirmed this PR has reasonable code coverage. You can run `npm run test:coverage` to see your progress.
  - [X] Confirmed this PR has unit test files
  - [X] Ran `npm run test`, made sure all tests are passing
- [X] Confirmed relevant documentation has been updated/added.

## Reviewer Checklist
_The reviewer of the PR should copy-paste this template into the review comments on review._

- [ ] Linting code actions have passed.
- [ ] Ran the code locally based on the test instructions.
[...]

PR #817, opened 2021-04-08
Bugfix small promo image override

JackHowa:
- use small promo block
- set a content api like `/local/2020/01/14/in-albania-age-old-traditions-and-mediterranean-beaches-on-the-cheap/`
- see the headline and image on publish
- then update the content override image `https://cloudfront-us-east-1.images.arcpublishing.com/corecomponents/QI6YYS4M7VAAXLMN6R6J6FFCQY.jpg`
- then publish
- previously, we wouldn't see the image override in the published version. all admin works as expected


JackHowa:
Closed in favor of #816

PR #825, opened 2021-04-13
TMEDIA-166 - Search content source encoding

matthewroach:
## Description

Encode the query string value being used to call the search API 

## Jira Ticket
- [TMEDIA-166](https://arcpublishing.atlassian.net/browse/TMEDIA-166)

## Test Steps

1. Checkout this branch `git checkout TMEDIA-166-search-content-source-encoding`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/search-content-source-block`
3. Using content debugger - http://localhost/pagebuilder/tools/debugger?_website=the-gazette&type=content-debugger&source=search-api
4. Using the following value - `vidèo` verify results are now returned
5. Verify you see a search result in the theme search results page - http://localhost/search/vid%C3%A8o/?_website=the-gazette

### Before

Search results not found -

TMEDIA-166-after

### After

Search results now found - 
TMEDIA-166-after

## Author Checklist
- [X] Confirmed all the test steps a reviewer will follow above are working.
- [X] Confirmed there are no linter errors. Please run `npm run lint` to check for errors. Often, `npm run lint:fix` will fix those errors and warnings.
  - [X] Ran `npm run test`, made sure all tests are passing
- [X] Confirmed relevant documentation has been updated/added.

## Reviewer Checklist

### DIFF
--- a/blocks/search-content-source-block/sources/search-api.js
+++ b/blocks/search-content-source-block/sources/search-api.js
@@ -5,7 +5,7 @@ export default {
   resolve(contentOptions) {
     const { query, page, 'arc-site': arcSite } = contentOptions;
     if (query) {
-      return `https://search.arcpublishing.com/search?&q=${query}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
+      return `https://search.arcpublishing.com/search?&q=${encodeURIComponent(query)}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
     }
     return '';
   },
--- a/blocks/search-content-source-block/sources/search-api.js
+++ b/blocks/search-content-source-block/sources/search-api.js
@@ -5,7 +5,7 @@ export default {
   resolve(contentOptions) {
     const { query, page, 'arc-site': arcSite } = contentOptions;
     if (query) {
-      return `https://search.arcpublishing.com/search?&q=${encodeURIComponent(query)}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
+      return `https://search.arcpublishing.com/search?&q=${encodeURIComponent(decodeURIComponent(query))}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
     }
     return '';
   },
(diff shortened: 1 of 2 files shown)


blocks/search-content-source-block/sources/search-api.js
```diff
@@ -5,7 +5,7 @@ export default {
   resolve(contentOptions) {
     const { query, page, 'arc-site': arcSite } = contentOptions;
     if (query) {
-      return `https://search.arcpublishing.com/search?&q=${query}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
+      return `https://search.arcpublishing.com/search?&q=${encodeURIComponent(decodeURIComponent(query))}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
```
JackHowa, line 8: not sure I understand why the query is being decoded first if we're trying to encode it?

JackHowa, line 8:
```suggestion
      return `https://search.arcpublishing.com/search?&q=${encodeURIComponent(query)}&page=${page || 1}${arcSite ? `&website_id=${arcSite}` : ''}${SEARCH_KEY ? `&key=${SEARCH_KEY}` : ''}`;
```

blocks/search-content-source-block/sources/search-api.test.js
```diff
@@ -16,6 +16,14 @@ describe('the search content source block', () => {
       });
       expect(url).toEqual('https://search.arcpublishing.com/search?&q=test&page=1&website_id=the-sun&key=ABCDEF');
     });
+
+    it('should encode the query value', () => {
+      const url = contentSource.resolve({
+        query: 'vidèo',
```
JackHowa, line 22: nice test, helpful to understand an example where this is used

JackHowa (changes requested):
I might be missing the reason for decodeURI. But I was getting the same results without decode testing it locally. Likely not a big deal but curious

blocks/search-content-source-block/sources/search-api.js
matthewroach: We have to encode the query to account for the accent characters being sent to the search API.. But in some cases the query is already encoded, so to account for that we decode and encode it. To test this you can remove the `decodeURIComponent` part of this and then load the search page - http://localhost/search/vid%C3%A8o/?_website=the-gazette and you'll see it's double encoding the query value and returning no results

JackHowa: `But in some cases the query is already encoded` didn't know this bit

JackHowa (approved):
Approved. Didn't realize there was an edge case where the query is already encoded coming in

PR #831, opened 2021-04-14
ci: Add clarifying time period and include rc and canary publish

JackHowa:
- clarify date separation
- add to rc and beta


matthewroach (approved):
Thanks @JackHowa, nice work!

PR #833, opened 2021-04-15
Tmedia 201 spa output type

JackHowa:
- published to latest the two new blocks to help linking 
- added a countdown block to check if react state can persist and use static
- Created a new spa output type


matthewroach:
Jason has an updated Fusion release that allows for local linking without the need to publish blocks, did you try his Fusion build?

JackHowa:
will keep this open for reference. but the spa output type seems to be working as expected for default output type

JackHowa:
spa = true is the really only big need for default output type in the future

PR #836, opened 2021-04-19
TMEDIA-194 - Ad Block reserve space

matthewroach:
## Description

Add new custom field to ad-block which defaults to true - which ads logic to reserve space for the ad to fill to avoid layout shifting.

## Jira Ticket
- [TMEDIA-194](https://arcpublishing.atlassian.net/browse/TMEDIA-194)

## Acceptance Criteria
1. A new boolean Custom Field is added to the Ad Block called Reserve space for ad? 
    * If the value is true then the block should reserve enough space for the ad to fill in on client side load
    * If the value is false or not present then no change in behavior needs to happen, it behaves as it does now (not reserving space) 
    * The field should default to true
2. If there are multiple heights requested for the ad, it should reserve the smaller height. It should reserve the relevant height by breakpoint.

## Test Steps

1. Checkout this branch `git checkout TMEDIA-194-reserve-ad-space`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/ads-block`
3. Visit a page with ads on and verify page loads with space for the ad to fill, and then ads are loaded and do not cause layout shifting - http://localhost/pf/adstesting/?_website=the-gazette
4. Article page with Ads - http://localhost/2019/09/17/global-kitchen-sink-article/?_website=the-gazette
5. Edit the ads testing page to disabled reserve space and verify layout shifting is back - Not custom fields for existing blocks on a page do inherit default values.

### Before

Ads caused lot of layout shifting

TMEDIA-194-before

### After

Reduced layout shifting - still see some layout shift based on what ad is being served in slot that have multiple sizes allocated to them

TMEDIA-194-after

## Author Checklist
- [X] Confirmed all the test steps a reviewer will follow above are working.
- [X] Confirmed there are no linter errors. Please run `npm run lint` to check for errors. Often, `npm run lint:fix` will fix those errors and warnings.
  - [X] Ran `npm run test`, made sure all tests are passing

## Reviewer Checklist

### DIFF
--- a/blocks/ads-block/features/ads/default.jsx
+++ b/blocks/ads-block/features/ads/default.jsx
@@ -77,6 +77,11 @@ const ArcAd = (props) => {
   // Height is + 17px to account line-height of advertisment string of 17px;
   const heightWithAdjustments = parseInt(height, 10) + ((displayAdLabel) ? 17 : 0);
 
+  const sizing = {
+    width: `${width}px`,
+    minHeight: reserveSpace ? `${heightWithAdjustments}px` : null,
+  };
+
   return (
     <StyledAdUnit
       id={`arcad-feature-${instanceId}`}
@@ -84,7 +89,7 @@ const ArcAd = (props) => {
       adLabel={siteProperties?.advertisementLabel || 'ADVERTISEMENT'}
       displayAdLabel={!isAdmin && displayAdLabel && !isAMP()}
     >
-      <div className="arcad-container" style={reserveSpace ? { minHeight: `${heightWithAdjustments}px`, width: `${width}px` } : {}}>
+      <div className="arcad-container" style={sizing}>
         {!isAdmin && !isAMP() && (
           <LazyLoad enabled={lazyLoad}>
             <AdUnit
--- a/blocks/ads-block/features/ads/default.jsx
+++ b/blocks/ads-block/features/ads/default.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable react-hooks/rules-of-hooks */
 import React, { useState } from 'react';
 import PropTypes from '@arc-fusion/prop-types';
 import styled from 'styled-components';
--- a/blocks/ads-block/features/ads/default.jsx
+++ b/blocks/ads-block/features/ads/default.jsx
@@ -74,16 +74,17 @@ const ArcAd = (props) => {
   );
 
   const [width, height] = config.adClass ? config.adClass.split('x') : [];
+  // Height is + 17px to account line-height of advertisment string of 17px;
+  const heightWithAdjustments = parseInt(height, 10) + ((displayAdLabel) ? 17 : 0);
 
   return (
     <StyledAdUnit
-      id={`arcad_feature-${instanceId}`}
-      className="arcad_feature"
+      id={`arcad-feature-${instanceId}`}
+      className="arcad-feature"
       adLabel={siteProperties?.advertisementLabel || 'ADVERTISEMENT'}
       displayAdLabel={!isAdmin && displayAdLabel && !isAMP()}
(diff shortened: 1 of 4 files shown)


blocks/ads-block/features/ads/default.jsx
```diff
@@ -33,7 +33,6 @@ const StyledAdUnit = styled.div`
 `;
 
 const ArcAd = (props) => {
-  if (typeof window === 'undefined') return null;
```
JackHowa, line 36: was this change deliberate?

blocks/ads-block/features/ads/default.test.jsx
```diff
@@ -96,14 +97,38 @@ describe('<ArcAd>', () => {
...
-        const adUnitEl = wrapper.find('.arcad_feature LazyLoad AdUnit');
+        const adUnitEl = wrapper.find('.arcad-feature LazyLoad AdUnit');
         expect(adUnitEl).toHaveLength(1);
         expect(typeof adUnitEl.prop('adConfig')).toEqual('object');
         expect(typeof adUnitEl.prop('featureConfig')).toEqual('object');
       });
     });
   });
 
+  describe('Reserve Space', () => {
```
JackHowa, line 108: looks good, nice work

blocks/shared-styles/_children/promo-date/index.test.jsx
```diff
@@ -15,7 +15,7 @@ jest.mock('fusion:properties', () => jest.fn(() => (
 )));
 
 jest.mock('@wpmedia/engine-theme-sdk', () => ({
-  localizeDateTime: jest.fn(() => new Date().toDateString()),
+  localizeDateTime: jest.fn((x) => new Date(x).toDateString()),
```
JackHowa, line 18: ew, I'm not sure how I feel about this mock. good job improving it though. I might prefer just to pass through a fake date?

JackHowa (dismissed):
I think we need to keep the window check if I'm not mistaken. the rest looks pretty solid, especially with min-content

blocks/ads-block/features/ads/default.jsx
matthewroach: Needs to return from the server so we can place the element on the page to reserve the space

JackHowa: wait, we need this

JackHowa: yep just missed your comment. was thinking about this more heh

vinnie-olsen: Just for context, I originally added the window check to prevent the ads code from executing server-side and rendering a blank ad placeholder element. That being said, I was planning on refactoring this because I think it breaks the rules of hooks since it returns before the `useState` calls (but then I left the team and never got a chance to actually do that). 

I don't _think_ removing it should cause any issues since I don't think the `arcads` library should actually make ad requests when server-side. I'm not 100% sure on that part though. There might need to be a check to prevent `arcads` from being initialized & registering ad units when server-side if not preventing the `ads-block` from rendering server-side. Not completely sure without testing that though.

JackHowa: ok cool, we'll move forward with removing this. @matthewroach can you remove the rule of hooks ignore in the first line

JackHowa (dismissed):

blocks/ads-block/features/ads/default.jsx
JackHowa: thanks for the clarification @vinnie-olsen !

JackHowa (approved):

blocks/ads-block/features/ads/default.jsx
vinnie-olsen: no problem @JackHowa 🙂

PR #837, opened 2021-04-19
chore: Use more stable jest mock for time

JackHowa:
- fails github action https://github.com/WPMedia/fusion-news-theme-blocks/runs/2381249230?check_suite_focus=true#step:6:243


matthewroach:
Added in #836

PR #844, opened 2021-04-23
Update stories to be organized into folders

matthewroach:
Just a little tidy up to organize the stories a little better

Storybook-organised


JackHowa (approved):
good idea

PR #862, opened 2021-05-10
Update storybook to v6

matthewroach:


JackHowa:
excited for this ... will review

.storybook/.babelrc
```diff
@@ -1,10 +1,6 @@
 {
   "plugins": [
-    [
-      "@babel/plugin-proposal-decorators",
-      {
-        "legacy": true
-      }
-    ]
+    ["@babel/plugin-proposal-private-property-in-object", { "loose": true }],
```
JackHowa, line 3: did you add this plugin?

package.json
```diff
@@ -15,7 +15,7 @@
     "lint:styles": "stylelint '**/*.(scss|jsx)' --formatter verbose",
     "lint:styles:fix": "stylelint '**/*.scss' --fix",
     "lint:fix": "npm run lint -- --fix",
-    "storybook": "start-storybook",
+    "storybook": "start-storybook -p 60001",
```
JackHowa, line 18: why does it need to be on this port?

JackHowa (dismissed):
works as expected! looks great

JackHowa:
closes https://github.com/WPMedia/fusion-news-theme-blocks/issues/487

JackHowa:
looks like there's some substantive differences in the review stage? https://www.chromatic.com/test?appId=5f91e4721ccaba0022a10782&id=6098f4a784ad4f003bcdae8e old diff?

package.json
matthewroach: Not needed per say, but as we have more storybooks having them on a specific port allows better linking

.storybook/.babelrc
matthewroach: Starting storybook threw a load of warnings suggesting to add this.

package.json
JackHowa: gotcha no worries, thought it may have been accidental

matthewroach:
Not sure about the chromatic tests as they shouldn't have changed

.storybook/.babelrc
JackHowa: huh interesting we didn't need to add it to package.json. I guess it must be a nested dep

matthewroach:
Fixed the storybook layout issues @JackHowa in v6 they have a layout parameter that has a different default to pre v6 - https://storybook.js.org/docs/react/configure/story-layout

JackHowa:
> Fixed the storybook layout issues @JackHowa in v6 they have a layout parameter that has a different default to pre v6 - https://storybook.js.org/docs/react/configure/story-layout

nice good find. didn't know about this one

JackHowa (approved):
good work! excited this is working

PR #872, opened 2021-05-18
TMEDIA-218 - Lokalise i18n

matthewroach:
## Description

Better locale file management as we work towards using Lokalise for our translations.

* All language key/values to be stored in {locale}.json files within the repository - which is also stored and sync'd in Lokalise
* Added a node script to generate each blocks intl.json file 
    * Will look at automation for this in future

{locale}.json file structure
```
{
  "test-block": {
    "translation-key": "Translation key value"
  },
  "another-blockname": {
    "translation-key": "Translation key value"
  }
}
```

## Jira Ticket
- [TMEDIA-218](https://arcpublishing.atlassian.net/browse/TMEDIA-218)

## Reviewer Checklist


matthewroach:
@JackHowa I updated canary on this PR, ready for review again

locale/scripts/generate-intl.js
```diff
@@ -0,0 +1,39 @@
...
+    keys.forEach((item) => {
+      if (!blockOutputMap[item]) {
+        blockOutputMap[item] = {};
+      }
+
+      blockOutputMap[item][l] = fileMappings[l][key][item];
+    });
+  });
+
+  fs.writeFileSync(`./blocks/${key}/intl.json`, JSON.stringify(blockOutputMap));
```
JackHowa: this is somewhat cosmetic, but it'd make validating the output json easier for me. right now, the output json isn't formatted like the previous one. I can spot-check it but hard to validate properly https://stackoverflow.com/a/3515761/7491536

JackHowa:
```suggestion
  fs.writeFileSync(`./blocks/${key}/intl.json`, JSON.stringify(blockOutputMap, null, "\t"));
```

package.json
```diff
@@ -21,6 +21,7 @@
     "latest-stable-parity": "lerna exec --no-bail --parallel -- npm dist-tag add \\${LERNA_PACKAGE_NAME}@$(node -e \"console.log(require('./lerna.json').version)\") stable",
     "build-storybook": "build-storybook -c .storybook -o news-theme-blocks",
     "test:coverage": "jest --coverage",
+    "generate:intl": "node locale/scripts/generate-intl.js",
```
JackHowa, line 24: nice. are there docs for this somewhere?

locale/scripts/generate-intl.js
```diff
@@ -0,0 +1,39 @@
...
+localeFiles.forEach((file) => {
+  const lang = path.basename(file).split('.json')[0];
+  supportedLocales.push(lang);
+  fileMappings[lang] = require(resolve(file));
+});
+
+// JSON is strucutred with first key being the block name
+const blockNames = Object.keys(fileMappings[defaultLang]);
+
+blockNames.forEach((key) => {
```
JackHowa, line 22: looks good

JackHowa (changes requested):
Have a two small suggestions. but they are not necessarily big issues. since this is a spike ticket, I want to make sure we have a good base

package.json
matthewroach: Just as part of the dropbox doc I put together.. Certainly will need to document this within confluence on how we do translations

blocks/global-phrases-block/intl.json
```diff
@@ -1,48 +1,62 @@
...
+		"no": "Stopp",
+		"sv": "Pausa"
+	},
+	"global.gallery-autoplay-label-start": {
+		"de": "Start automatic slide show",
+		"en": "Start automatic slide show",
+		"es": "Start automatic slide show",
+		"fr": "Start automatic slide show",
+		"ja": "Start automatic slide show",
+		"ko": "Start automatic slide show",
```
JackHowa, line 48: looks like this wasn't translated?

JackHowa: oh, this is deliberate, right?

package.json
JackHowa: ah gotcha

JackHowa (approved):
Looks good! thanks for adding formatting for readability

blocks/global-phrases-block/intl.json
matthewroach: Correct, this is now generating the english for missing translations

PR #879, opened 2021-05-20
Update workflow name

matthewroach:
Workflows name was missing "publish"


matthewroach:
This is incorrect.. Issue is the actions need to be on the "default" branch which is stable for this repo

PR #904, opened 2021-06-15
TMEDIA-238 - Add intl.json to package.json files array

matthewroach:
Update package.json files to ensure blocks that now have `intl.json` files are published


JackHowa (approved):
good good, I always forget this too

PR #906, opened 2021-06-15
Fix use extract image helper numbered list

JackHowa:
- fix tests 
- see similar fix https://github.com/WPMedia/fusion-news-theme-blocks/pull/903/files#diff-830f05ceb038c74d8ffef5b6d4af88f446767da639f750a9d3b568560ddca3d9R6


matthewroach:
This will be fixed in https://github.com/WPMedia/fusion-news-theme-blocks/pull/913

PR #937, opened 2021-07-01
feat: Toggle arc static enabling, simplify render logic

JackHowa:
- use with https://github.com/WPMedia/engine-theme-sdk/pull/379 
- Continue rendering logic for arc-static and loading="lazy" and user-enabled lazyLoad
- npx fusion start -f -l @wpmedia/card-list-block

## lazy load off 

create block with many stories. lazyLoad off
- should all render with static without js enabled

- images only loading once 
Screen Shot 2021-07-01 at 14 26 23

## lazy load on 

create block with many stories. lazyLoad on

- add many brs to an html box to lazy load the whole block
- see the requests only begin as you scroll the block into viewport 
Screen Shot 2021-07-01 at 14 47 02
Screen Shot 2021-07-01 at 14 46 23
Screen Shot 2021-07-01 at 14 45 25

- when js is disabled, images should not be called nor render when the content is not in view 
- when you scroll down, with js disabled, and with lazy load on, the images should be displayed as normal

### DIFF
--- a/blocks/article-body-block/chains/article-body/default.jsx
+++ b/blocks/article-body-block/chains/article-body/default.jsx
@@ -93,6 +93,7 @@ function parseArticleItem(item, index, arcSite, phrases, id, customFields) {
             largeHeight={0}
             breakpoints={getProperties(arcSite)?.breakpoints}
             resizerURL={getProperties(arcSite)?.resizerURL}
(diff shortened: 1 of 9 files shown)


JackHowa:
closed by engine theme sdk pr

PR #959, opened 2021-07-07
TMEDIA-289 - Review Results List for React - Optimization

nschubach:
## Description
Rewrite block to use functional components.  Rewrote tests using `@testing-library/react`

## Jira Ticket
- [TMEDIA-289](https://arcpublishing.atlassian.net/browse/TMEDIA-289)

## Acceptance Criteria
Optimize React Render methods

## Test Steps

0. Run `npm i` for additional packages around testing
1. Checkout this branch `git checkout TMEDIA-289-result-list-refactor`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/results-list-block`
3. Create a new page and add the results-list-block to it
4. Configure the results-list-block as follows:
image
5. You should see it render the `/sara-carothers` articles 5 at a time.
6. Configure the results-list-block as follows:
image
7. You should see the homepage articles starting from the third article. (verify this by setting from to 0 and see the third)
8. Configure the following values and validate no button (I only have two stories for this, if you have more, increase size):
image
9. Page should preload 5 additional items, but not render them.  You can check this by viewing the network tab in debug and counting the number of items returned from the initial request.  Any press of the `See More` button will only pull the number of items equal to your size.
image
image
image

### Before
Pages would load content as requested, causing a slight delay to the user
Render passes (first and after hitting see more per @JackHowa's recommendation):
image
image
Lighthouse Result:
image

### After
Pages should show additional content immediately and load more in the background
Render passes:
image
image
Lighthouse Result:
image

N/A

## Author Checklist

## Reviewer Checklist


matthewroach (changes requested):
With such a big change set I've been testing it as a whole and these are the following issues/changes that differ based on the changes made

* Lazy Load option is still requesting the data on the server side
* Results list is never rendered on the server side, it is now completely client side rendered
* The initial render is set the page's focus to the first element of the results list and breaking the focus order - Focus should only be manually set after the user clicks see more
* Focus isn't set on the first new item to be rendered after clicking see more button when you have an offset value set
* The see more button is delayed in showing after the next set of results are shown

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,326 @@
...
+        const feedOffset = requestedOffset === configuredOffset
+          ? configuredOffset
+          : requestedOffset + configuredSize;
+        return { feedOffset, feedSize };
+      }
+      case 'content-api-collections': {
+        const size = requestedOffset === configuredOffset
+          ? configuredSize * 2
+          : configuredSize;
+        const from = requestedOffset === configuredOffset
```
matthewroach: Each `case` block works out the same size and offset. The only difference is the object retuned keys. We could reduce the duplication and just return in correct formats in the `case` blocks

nschubach:
fixed - Lazy Load option is still requesting the data on the server side
fixed - The initial render is set the page's focus to the first element of the results list and breaking the focus order - Focus should only be manually set after the user clicks see more
fixed - Focus isn't set on the first new item to be rendered after clicking see more button when you have an offset value set
fixed - Results list is never rendered on the server side, it is now completely client side rendered
investigating - The see more button is delayed in showing after the next set of results are shown

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: fixed in the latest

nschubach:
I can update the visibility logic of the button to make it remain on screen if a current request is active, but that shifts the "change" to make the button remain on screen until it finds out if there is more content and makes it vanish at the end of the list.

Is this preferable?  Maybe it should disable...

matthewroach:
The button being removed and then shown quickly was something I noticed as different between current and the updates in this pull request.

nschubach:
This changes the 'See More' button visibility functionality and introduces measures to prevent re-renders.

blocks/results-list-block/features/results-list/default.jsx
```diff
@@ -1,407 +1,30 @@
...
-          <Results />
-        </HeadingSection>
-      </LazyLoad>
-    );
-  }
-}
+import Results from './results';
+
+const ResultsList = () => {
+  const { customFields, isAdmin } = useFusionContext();
```
matthewroach: `customFields` is a prop of features in Fusion. Getting them from a fusion hook seems like an anti pattern, and also something that is not consistent with other blocks

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+}));
+
+const Results = () => {
+  const {
+    arcSite,
+    contextPath,
+    customFields,
+    deployment,
+    isAdmin,
+  } = useFusionContext();
```
matthewroach: `useFusionContext` is used in the parent component - would it not be more performant to get the items needed in the parent and pass them in as props?

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+    },
+  } = customFields;
+
+  const [{
+    breakpoints,
+    fallbackImage,
+    locale,
+    primaryLogoAlt,
+    resizerURL,
+  }] = useState(getProperties(arcSite) || {});
```
matthewroach: Lifting this to the parent and passing as prop would remove the need for `useState`

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+
+  const [imageProps] = useState({
```
matthewroach: Lifting this to the parent and passing as prop would remove the need for `useState`

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+          ref={listItemRefs.current[element._id]}
+          arcSite={arcSite}
+          element={element}
+          promoElements={promoElements}
+          imageProps={imageProps}
+          targetFallbackImage={targetFallbackImage}
+          placeholderResizedImageOptions={placeholderResizedImageOptions}
+        />
+      ))}
+      {(viewableElements.length < resultList?.content_elements.length
```
matthewroach: See more button is showing even when the results list is shorter than the custom field size value.
Example 
content source - story-feed-query 
quey - matthew
size - 50

There are only 2 results for query `matthew` - but See More button is still shown

blocks/results-list-block/features/results-list/default.jsx
nschubach: I thought it was more dirty to grab different values from various places, but I'll change it back.  Makes testing easier...

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: I figured this was cached... thereby making the "performance" nearly minuscule but I can change it.  Makes testing easier...

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: Makes sense with the other changes.

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: Makes sense with the other changes.

matthewroach:
Do we have before and after performance metrics on the changes in the pull request? For such a drastic change in approach of the block it would be something worth while finding out if we are making things better or worse

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: Resolved in 5286b9a

blocks/results-list-block/features/results-list/default.jsx
```diff
@@ -1,407 +1,50 @@
...
+    breakpoints: fusionProperties.breakpoints,
+    resizerURL: fusionProperties.resizerURL,
+  };
+
+  return (
+    <LazyLoad enabled={lazyLoad && !isAdmin}>
+      <HeadingSection>
+        <Results
+          customFields={customFields}
+          fusionContext={fusionContext}
```
matthewroach: You are passing the entire fusionContent object to the child. This a very large object that the child component only needs a handful of items from, we should only pass what is needed.

blocks/results-list-block/features/results-list/default.jsx
```diff
@@ -1,407 +1,50 @@
...
+          fusionProperties={fusionProperties}
```
matthewroach: Same as above, we should only pass down what is needed, this object isn't as big as the content by could become very large for some clients

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,347 @@
...
+    deployment,
+    isAdmin,
+  } = fusionContext;
+
+  const {
+    fallbackImage,
+    locale,
+  } = fusionProperties;
+
+  const [configuredOffset] = useState(
```
matthewroach: What's the reason to use state to store everything in?

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,347 @@
...
+    parseInt(contentConfigValues.offset, 10)
+    || parseInt(contentConfigValues.feedOffset, 10)
+    || parseInt(contentConfigValues.from, 10)
+    || 0,
+  );
+  const [configuredSize] = useState(
+    parseInt(contentConfigValues.size, 10)
+    || parseInt(contentConfigValues.feedSize, 10)
+    || 10,
+  );
```
JackHowa: is there a reason why these are not set as variables without `useState`? They appear to help derive state elsewhere

```suggestion
  const configuredOffset = parseInt(contentConfigValues.offset, 10)
    || parseInt(contentConfigValues.feedOffset, 10)
    || parseInt(contentConfigValues.from, 10)
    || 0;

  const configuredSize = parseInt(contentConfigValues.size, 10)
    || parseInt(contentConfigValues.feedSize, 10)
    || 10;
```

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,347 @@
...
+  background-color: ${(props) => props.primaryColor};
+  font-family: ${(props) => props.primaryFont};
+
+  &:not(:disabled):not(.disabled):active:hover,
+  &:not(:disabled):not(.disabled):hover:hover {
+    background-color: ${(props) => props.primaryColor};
+  }
+`;
+
+const ResultsItem = React.memo(React.forwardRef(({
```
matthewroach: Pulling this out to it's own file would make this file a bit more readable, and possibly allow us to in the future utilize this component in other list blocks

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,347 @@
...
+    query: { raw_image_url: targetFallbackImage, respect_aspect_ratio: true },
+  });
+
+  const [isServerSideLazy] = useState(lazyLoad && isServerSide() && !isAdmin);
+
+  const serviceQueryPage = useCallback((requestedOffset = 0) => {
+    /*
+      This sets up a window view of the data starting from the initial index
+      to *twice* the size.  When clicking on the more button, we will render
+      the next size worth of items and load the next size from the server.
```
JackHowa, line 49: this is different behavior than what we had initially, right? Assuming so, this decision to load more data than is necessary will have the benefit of faster functionality later. But there's the tradeoff that this will slow down the first render. Was this part of the intention of the ticket?

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,347 @@
...
+      ? deployment(`${contextPath}/${fallbackImage}`)
+      : fallbackImage,
+  );
+
+  const placeholderResizedImageOptions = useContent({
+    source: !targetFallbackImage.includes('/resources/') ? 'resize-image-api' : null,
+    query: { raw_image_url: targetFallbackImage, respect_aspect_ratio: true },
+  });
+
+  const [isServerSideLazy] = useState(lazyLoad && isServerSide() && !isAdmin);
```
JackHowa: this also seems like a candidate for refactoring. not sure why this is necessary to save in state. I'd recommend this blog post on when and when not to sync state https://kentcdodds.com/blog/dont-sync-state-derive-it

JackHowa:
```suggestion
  const isServerSideLazy= lazyLoad && isServerSide() && !isAdmin;
```

JackHowa: if you look elsewhere in the code, this similar functionality is not set as react state

blocks/results-list-block/features/results-list/results/index.jsx
```diff
@@ -0,0 +1,347 @@
...
+    if (focalElement?.current) {
+      const focusLink = focalElement.current.querySelector('a:not([aria-hidden])');
+      if (focusLink) {
+        focusLink.focus();
+      }
+    }
+  }, [focalElement]);
+
+  const [promoElements] = useState(resolveDefaultPromoElements(customFields));
+  const [phrases] = useState(getTranslatedPhrases(locale || 'en'));
```
JackHowa: similarly here

nschubach:
Added some Performance metrics... lighthouse I think it in margin of error... react tools shows difference of upfront loading on the new code.

JackHowa (changes requested):
First of, I applaud you for diving into this functionality. Using useMemo and ref can be a powerful tool for reducing render cycles and unnecessary re-rendering. I think the the `See More` functionality is less glitchy now. But I'm not sure if that was the intent of the ticket. When I looked at this ticket myself, I looked at reducing render cycles and re-rendering. The react profiler was helpful for that. This stuff can be hard but I'll try to look at it on a line-by-line basis.

Looks like there's a few places where there's unnecessary useState usages. I'd recommend looking at when to use useState vs setting a plain old variable. I included a good resource in one of the comments. I removed the flagged `useState`s and pushed an alternate [branch](https://github.com/WPMedia/fusion-news-theme-blocks/tree/TMEDIA-289-result-list-refactor-derive-state) with seemingly no ill effects in my testing. 

I see that useMemo was used. I know the ticket was about reducing unnecessary render cycles, which can be tough. Overall, I found one more render cycles in this feature branch (4) than in canary (3). It also seemed like some content was re-rendered in the optimization feature branch. Were you able to see this in the react profiler? Update: based on your screenshots, they don't seem to match mine? I'm curious what behavior you used to see that

Another goal of the ticket was to reduce dom nodes and improve ligthouse score. I was not able to find a consistent difference between this branch and canary. Was there something that was done to improve the performance score? 

It's hard for me to compare on a line-by-line basis as there have been file renaming. And I'm not sure if we have adequate testing resources (which have been changed in this pr) to start with, to do a refactor of this magnitude. I like to keep my tests constant if I make a sizable refactor to a component. 

canary

Screen Shot 2021-07-08 at 15 37 20

this branch 

Screen Shot 2021-07-08 at 15 22 18
Screen Shot 2021-07-08 at 15 22 05

JackHowa:
what do you mean by this? @nschubach 

> react tools shows difference of upfront loading on the new code.

blocks/results-list-block/features/results-list/results/helpers.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+const reduceResultList = (state = defaultResultList, action) => {
+  const { type, data } = action;
+
+  if (!data) { return state; }
+
+  const oldItemIds = state.content_elements.map((element) => element._id);
+  const uniqueNewItems = data.content_elements
+    .filter((element) => !oldItemIds.includes(element._id));
```
JackHowa, line 12: is there fallback if there's no content_elements or an _id. I don't remember this functionality from before

blocks/results-list-block/package.json
```diff
@@ -36,5 +36,7 @@
     "@wpmedia/shared-styles": "*",
     "styled-components": "^4.4.0"
   },
-  "gitHead": "83fb8d6685958d85ddbfcfbf01d9a5c864cd6c95"
+  "gitHead": "83fb8d6685958d85ddbfcfbf01d9a5c864cd6c95",
+  "devDependencies": {},
+  "dependencies": {}
```
JackHowa: dev and dependencies can be removed

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: CC: @matthewroach : This is the main "win" for useState.  These functions (at least one of them) returns an object.  That object is a new object each time the render is called, so it causes the memoized components (and useEffects) to re-render every time even thought he values within don't change.  Since the phrases and the promoElements are not likely to change between renders, making them "state" is a way to prevent re-rendering.

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: I explained more in the comment to Jack below, but it's more to prevent re-renders. I don't need them on the simple values and am checking in code to revert some of that useState.

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: I was hoping you didn't notice that.  I definitely thought about this... I suppose since this story is carried on, I will undertake that and update/make the appropriate tests for it.

blocks/results-list-block/features/results-list/default.jsx
nschubach: resolving

blocks/results-list-block/features/results-list/default.jsx
nschubach: resolving

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: I'm removing those states... it was more of a knee jerk reaction to new items causing a re-render and I (wrongly) felt under the gun to get a working state quickly.

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: Heh, this was the initial concern for me on the ticket.  Previously, the feeds that did not have a `next` property would "preload" the next set of results to determine if the see more button should be shown, then they were thrown away.  So I thought it would be a better use experience and less wasteful to preload the next set of results so when the user hit the button there was no delay in showing them the results and we could load the next set in the background.

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: Yep, see above.

blocks/results-list-block/features/results-list/results/helpers.js
nschubach: You mean a fallback for elements without an _id?  From what I can tell, that's not possible?

nschubach: As far as functionality.  It's just me being paranoid about duplicate records.

blocks/results-list-block/package.json
nschubach: np.  they were added when I added the testing to this module, but I had to move it up to the root package.json.

blocks/results-list-block/features/results-list/results/index.jsx
JackHowa: "Yep", you're saying you see this elsewhere in the code? 
Screen Shot 2021-07-08 at 16 21 45

nschubach: sorry, "yep" I am removing the useState.  Sorry for the confusion.

nschubach:
> what do you mean by this? @nschubach
> 
> > react tools shows difference of upfront loading on the new code.

All I meant by that is, I was seeing the original code initially load an empty element and the changes and then rendered the list after the constructor fetch.  Interested in how you got only 3-4 renders though.

This was a 10 size, author feed render with new code:
image

Compared to the same page in canary:
image

The initial render in canary is a very lean ResultList:
image

blocks/results-list-block/features/results-list/results/index.jsx
JackHowa: so I'm not seeing any effect of phrases on clicking 'read more'. I have been seeing promoElements benefits in usestate for preventing re-renders. it looks like you're using the ref there. Couldn't we use a stable key for this to prevent react from re-rendering the preexisting list items? I can send you the json for the react profiler 

Screen Shot 2021-07-09 at 10 28 38

JackHowa: looks like each item is being passed all promoElements?

JackHowa: I was able to not useState for promoElements if I didn't pass each item all elements. I separated out the individual props of showDate, etc

https://github.com/WPMedia/fusion-news-theme-blocks/pull/967/files

Screen Shot 2021-07-09 at 10 38 36

blocks/results-list-block/features/results-list/results/index.jsx
JackHowa: I'm not able to replicate the reason for dismissing this. I was able to replicate this suggestion helping https://github.com/WPMedia/fusion-news-theme-blocks/pull/967/files

JackHowa:
@nschubach Sounds like the 23 renders you were seeing the screenshot is from the version of engine theme sdk from a couple weeks ago. let me know if you can replicate faster initial load and fewer render cycles. I'm not sure if this ticket would be best if it was refocused on improving the 'read more' functionality while maintaining the class component and the previous tests. Like I said, it's hard to evaluate a rewrite without the preexisting tests. Also, from my testing, the initial load is longer and with more more render cycles than canary. 

If this was me, I might start over and refocus the ticket on 'See more' functionality. Always good to measure performance after a refactor and evaluate if it was worthwhile. In my branch, with some of the suggestions, I didn't see huge improvements either. (cc @RobinGiannattasio )

RobinGiannattasio:
Hey all! Catching up a bit on this PR - it seems like maybe the best next step is to regroup on Monday with a meeting with everyone who is not on vacation (ENJOY IT MATTHEW) so we can more easily sync on the goal of the ticket (improved performance via metrics) vs the other findings here (the somewhat less than ideal behavior of the see more button originally). I'll send out a meeting invite.

I think until then, enjoy trivia in the next few minutes, and focus you work on other tickets to help reset and clear the brains :D

blocks/results-list-block/features/results-list/results/helpers.js
JackHowa: ok this is new but I don't think it could necessarily be an issue

blocks/results-list-block/features/results-list/results/index.jsx
JackHowa: so this is resolved, right?

JackHowa: no, it doesn't seem like it is

nschubach: I can if you would like.  It's going to involve pulling some tests and this into their own files... maybe an hour?

blocks/results-list-block/features/results-list/results/index.jsx
JackHowa: was this suggestion based on previous code style addressed @nschubach ?  We were talking about this being the pattern elsewhere

blocks/results-list-block/features/results-list/results/index.jsx
JackHowa: nice nice, so this was addressed

JackHowa (changes requested):
It looks like some of the default, fallback values are missing. I'm not sure if clients always have a content source defined for their blocks

in feature branch, I see a red warning and error boundary box when there's no content source: 

Screen Shot 2021-07-12 at 17 12 54

in canary, no red warning box with no content source: 

Screen Shot 2021-07-12 at 17 13 21

JackHowa:
Looks like there's also a memory leak of some sort that I'm seeing in the admin: 

with this content 
Screen Shot 2021-07-12 at 17 21 16

in canary branch, no memory leak warning 

Screen Shot 2021-07-12 at 17 20 51


in feature branch, I see memory leak warning 

Screen Shot 2021-07-12 at 17 21 44

nschubach:
I don't see the red warning on mine...
image

I'm not sure I changed anything that would fix that with moving that component into its own file.

JackHowa:
Screen Shot 2021-07-12 at 17 39 02

Huh I updated the schema but left the content source. then I clicked `update`. 

Were you able to see the memory leak warning?

nschubach:
I can see the memory warning, but I have not identified anything that would cause it yet.

I did fix the red box issue by putting in a nullish coalesce on the properties so it properly falls back to the default values that it should.
```parseInt(contentConfigValues?.offset, 10)```

Will be updating that here in a minute.

nschubach:
Interesting, the useContent is triggering that warning...
image

nschubach:
I might need to run that one up the line to fusion

Asked here: https://washpost.slack.com/archives/CB8ANFE04/p1626132348300300

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: done

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: done and committed

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: I addressed this by getting the config values form the parameters... regardless, I think it's a minor performance difference (if any since I'm still querying the context)

blocks/results-list-block/features/results-list/results/index.jsx
nschubach: I assume resolved since you close the branch and the code was removed?

nschubach:
Per Jason Young I have opened a ticket for the warning: https://arcpublishing.atlassian.net/browse/PEN-2018

RobinGiannattasio (approved):
Hey hey! Making sure this keeps moving - I went through the comments and it seems everything has been addressed. As per our sync earlier this week, let's move this through to QA. If any other issues come up, we can create new tickets to track the work.

Thanks all!

matthewroach (approved):

JackHowa (commented):
approved to go to qa by robin

JackHowa (approved):
Robin approved for qa

PR #962, opened 2021-07-07
perf: Update the preconnect and preload for the font

JackHowa:
- Biggest render-blocking asset in my pagespeed insights testing https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fcorecomponents-the-gazette-prod.cdn.arcpublishing.com%2Fall-blocks%2F

Screen Shot 2021-07-07 at 16 45 37

before: 

Screen Shot 2021-07-07 at 16 41 18

after: 

- no render-blocking js 

after-preconnect

side effects: 

- did not hurt or help cls 

before:

Screen Shot 2021-07-07 at 16 41 37

after:

after-preconnect-didnt-fix-cls

Based on https://css-tricks.com/how-to-load-fonts-in-a-way-that-fights-fout-and-makes-lighthouse-happy/

### DIFF
--- a/blocks/default-output-block/output-types/default.jsx
+++ b/blocks/default-output-block/output-types/default.jsx
@@ -52,19 +52,56 @@ const googleTagManagerNoScript = (gtmID) => {
   );
 };
 
+const optimalFontLoading = (fontUrl, index = '') => (
+  <>
+    <link
+      rel="preload"
+      as="style"
+      href={`${fontUrl}&display=swap`}
+    />
+    <link
+      rel="stylesheet"
+      key={fontUrl}
+      data-testid={`font-loading-url-${index}`}
+      href={`${fontUrl}&display=swap`}
+    />
+
+    <noscript>
+      <link
+        rel="stylesheet"
+        href={`${fontUrl}&display=swap`}
+      />
+    </noscript>
+  </>
+);
+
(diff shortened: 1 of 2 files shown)


blocks/default-output-block/output-types/default.jsx
```diff
@@ -52,19 +52,56 @@ const googleTagManagerNoScript = (gtmID) => {
...
+      href={`${fontUrl}&display=swap`}
+    />
+    <link
+      rel="stylesheet"
+      key={fontUrl}
+      data-testid={`font-loading-url-${index}`}
+      href={`${fontUrl}&display=swap`}
+    />
+
+    <noscript>
```
matthewroach: This `<noscript>` is doing nothing more or different to the `link` on line 62 - This technique is only used when you set stylesheets to be a print media by default and then switch them to all with onload technique

JackHowa: ah. that makes sense. I saw the onload technique but couldn't validate that switch working as expected. I'll remove the noscript

JackHowa: ah. that makes sense. I saw the onload technique but couldn't validate that switch working as expected. I'll remove the noscript

matthewroach (approved):

PR #967, opened 2021-07-09
Tmedia 289 result list refactor derive state

JackHowa:
- example refactor without re-rendering already rendered elements upon clicking read more button 

wait for full  initial load. then scroll down. then start profiling react. then click read more. then stop profiling

Screen Shot 2021-07-09 at 10 38 36


JackHowa:
closed as suggestions have been adopted #959

PR #970, opened 2021-07-13
Ci only notify themes group codeowners

JackHowa:
- @ WPMedia/arc-pagebuilder-engine should not be notified per every pr here 
Screen Shot 2021-07-13 at 09 33 07


JackHowa:
closed in favor of #972

PR #973, opened 2021-07-14
fix: Google font url optimization breaking preexisting fonts

JackHowa:
- Reverting change made in tmedia-85 
- font url displayed as before (not necessarily always google font url)

https://redirector.arcpublishing.com/alc/arc-products/themes/user-docs/theme-settings/#Integration-Settings
fontURL (either an array of strings or string) 
> URL to a CSS stylesheet which contains font definitions for primary/secondary fonts used on the site (if they are not system fonts). This is commonly a Google Fonts URL.

- `git checkout REVERT-TMEDIA-85-google-font-url-swap`
`npx fusion start -f -l @wpmedia/default-output-block`
- use the https://github.com/WPMedia/Fusion-News-Theme/pull/182 branch with the raw font url (non google)

before

Screen Shot 2021-07-14 at 09 20 31

after

Screen Shot 2021-07-14 at 09 19 43


RobinGiannattasio (approved):
This looks good to me. 

As mentioned on slack, I think the optimization is still good, but maybe the best place to handle it is when we create the blocks.json via theme settings.

PR #1020, opened 2021-08-02
Chore update arc proptypes

JackHowa:
Use arc proptypes not plain proptypes for necessary blocks


nschubach:
Several failing tests?
`TypeError: Cannot read property 'isRequired' of undefined`

```      1 | import React from 'react';
    > 2 | import PropTypes from '@arc-fusion/prop-types';
        | ^```

nschubach (changes requested):
Failing tests

PR #1053, opened 2021-08-18
TMEDIA-436 identity input field

nschubach:
## Description
Implement a standard form field component for handling the states of the Identity forms components.

## Jira Ticket
- [TMEDIA-436](https://arcpublishing.atlassian.net/browse/TMEDIA-436)

## Acceptance Criteria
* Error state shows on field if the email address format is not valid
* Error state shows on field if the password does not meet minimum requirements
* Error state shows if required field is left blank

## Test Steps
1. Checkout this branch `git checkout TMEDIA-436-identity-input-field`
2. Run `npm run storybook` to validate the Identity/Input Field states

## Effect Of Changes
image
image
image
image
image

## Author Checklist

## Reviewer Checklist


nschubach:
It looks like the twitter component has updated causing the article-body-block to update it's appearance...

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+    ...(minLength ? { minLength } : {}),
+    ...(maxLength ? { maxLength } : {}),
+    ...(name ? { name } : {}),
+    ...(validationPattern ? { pattern: validationPattern } : {}),
+    ...(placeholder ? { placeholder } : {}),
+    ...(required ? { required } : {}),
+  };
+
+  return (
+    <PrimaryFont as="label" className="xpmedia-form-field">
```
matthewroach: Using the implicit label technique is not the most accessible technique. Where possible we should always use the explicit label technique

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+      { label
+        ? <div className="xpmedia-form-field-label">{label}</div>
```
matthewroach: Label should be a required field - Not having a label will fail accessibility

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+      { tip || error
+        ? (
+          <div
+            className={[
+              'xpmedia-form-field-tip',
+              ...(error ? ['xpmedia-form-field-tip--error'] : []),
+            ].join(' ')}
+          >
+            { error
+              ? <span>{errorMessage}</span>
```
matthewroach: This error message will not alert assistive technologies it was shown. 

The element will need the `role` attribute applied as `role="alert"` and the DOM node will always need to be present even when no error message is shown

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+              : null}
+            { tip
+              ? <span>{tip}</span>
```
matthewroach, line 96: Switching to explicit label markup will require the tip and input to be updated to associate the tip text with the input when it is focus by assistive technologies

blocks/identity-block/components/FormInputField/index.story.jsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+export const textField = () => (
+  <FormInputField
+    label="Plain Text Field"
+    name="field1"
+  />
+);
+
+export const textFieldNoLabel = () => (
```
matthewroach: Not a valid story - an input with no label should not be allowed - this fails accessibility

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+  type = FIELD_TYPES.TEXT,
+  validationErrorMessage,
+  validationPattern,
+}) => {
+  const [error, setError] = useState(isDefaultError);
+  const [errorMessage, setErrorMessage] = useState(validationErrorMessage);
+
+  const handleBlur = (event) => {
+    setError(!!event.target.validationMessage);
+    setErrorMessage(event.target.validationMessage
```
matthewroach: `event.target.validationMessage` is native to the input. If no custom validation error message is passed we do not show an error message, is this desired behaviour?

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+      ? validationErrorMessage || event.target.validationMessage
+      : '');
+  };
+
+  const handleChange = (event) => {
```
matthewroach, line 43: The error message is only shown using the `onBlur` event - but if the error is shown and the user is updating the input the error message is "live"?

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+    if (error) {
+      setError(!!event.target.validationMessage);
+      setErrorMessage(event.target.validationMessage
+        ? validationErrorMessage || event.target.validationMessage
+        : '');
+    }
+    onChange(event);
```
matthewroach: Should we control what is passed back here? Should we only pass back the value of the input?

blocks/identity-block/components/FormInputField/index.story.jsx
```diff
@@ -0,0 +1,138 @@
+import React from 'react';
```
matthewroach, line 1: We want to make sure this story file is not published with the block

blocks/identity-block/components/FormInputField/index.test.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
matthewroach, line 1: This test file is not being picked up in the coverage checker - you can see that it's not listed in the table output at the bottom of the "Run tests with coverage" output https://github.com/WPMedia/fusion-news-theme-blocks/runs/3365681291?check_suite_focus=true

blocks/identity-block/components/FormInputField/styles.scss
```diff
@@ -0,0 +1,47 @@
...
+    color: $danger-color;
+  }
+}
+
+.xpmedia-form-field-tip {
+  color: $ui-medium-gray;
+
+  > span {
+    &::after {
+      content: ' ';
```
matthewroach: What's the need to add a blank space after the tip?

matthewroach (changes requested):

blocks/identity-block/components/FormInputField/index.story.jsx
nschubach: Ah, so this is why we've been putting the story file in the root?  I was curious about this.  Is there a way that we can resolve this or do I move all this into the root level and split it apart somehow in the story configuration?

blocks/identity-block/components/FormInputField/styles.scss
nschubach: Word spacing.  To separate the messages as if they were one string of text.  Technically it only needs to be in all but the last value but I figured a trailing style space would not be that big of an impact.

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Debated this... I considered only sending back the values, but I didn't know what we would need for future support.  I can definitely do that and come back later to add the additional needs later, but that might involve editing other forms that also use this field.  Figured it was safe to pass back the event object since it's a synthetic field and emulating the field onChange was the goal.

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Preference on this.  I didn't want the field to be updating error states while you were typing, but I thought it was alright to clear the error when the user completes the "puzzle".

I'm definitely open to debate on the functionality on this.  It could even change in the future when we get design input.

blocks/identity-block/components/FormInputField/index.test.jsx
nschubach: I'm not sure why... and would like to know.  I don't suppose you have any ideas?

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Yeah, if the field is not in error, I didn't want to show an error.

blocks/identity-block/components/FormInputField/index.jsx
nschubach: I considered this, but for different reasons.  I didn't have a good way to generate unique ids for this without leaning on lodash, context, or some more complex methods.  Will need to think on this.

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Makes sense.  Will adjust.

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Changing.

blocks/identity-block/components/FormInputField/index.test.jsx
matthewroach: I fixed the issue on canary, the folder wasn't part of the jest config

blocks/identity-block/components/FormInputField/index.story.jsx
matthewroach: We can add an npmignore to ignore story files just like we do for the shared styles folder

blocks/identity-block/components/FormInputField/index.jsx
matthewroach: We could either use the name prop, or add a prop for ID

blocks/identity-block/components/FormInputField/index.jsx
matthewroach: I was referring to the state that you can set a prop to make the input show as in an error state, but it does not show the error message only the input with a red border. If we can set the input to be in an error state on initial render we should also show the error message

blocks/identity-block/components/FormInputField/index.jsx
matthewroach: Makes sense - the way to make this future proof would be to send back an object like ```{ value: value}``` and that way we would be able to expand on the keys being passed back without breaking existing uses

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Changed

blocks/identity-block/components/FormInputField/index.jsx
nschubach: I made name required and used it in the id

blocks/identity-block/components/FormInputField/styles.scss
nschubach: I changed this to just add a space to the error message.

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+              'xpmedia-form-field-tip',
+              ...(error ? ['xpmedia-form-field-tip--error'] : []),
+            ].join(' ')}
+            id={infoId}
+          >
+            { error
+              ? <span role="alert">{`${errorMessage} `}</span>
+              : null}
+            { tip
+              ? <span role="tooltip">{tip}</span>
```
matthewroach: The role `tooltip` is being used incorrectly here - we can forgo this here

matthewroach (changes requested):
Errors are showing `undefined` in Chromatic and on Safari locally - https://www.chromatic.com/test?appId=5f91e4721ccaba0022a10782&id=611e7a5b2a53eb003a3f6c69

blocks/identity-block/components/FormInputField/index.jsx
nschubach: I updated how the errors are tracked using more of the native functionality so it shows up if you pass the now renamed flag `showDefaultError`

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Updated to do this instead of passing the event.

blocks/identity-block/components/FormInputField/index.jsx
nschubach: Removed

nschubach:
I fixed the `undefined` error state.  I overlooked a condition in the last update.

blocks/identity-block/components/FormInputField/index.story.jsx
nschubach: In the identity-block/components?
image

matthewroach: Yes, that should work

matthewroach (dismissed):
All looking good now @nschubach 

If we can just get the npmignore set up we should be good to approve this and get it merged

blocks/identity-block/components/FormInputField/index.story.jsx
matthewroach: To test you can run `npm pack` from the the directory `blocks/identity-block` and it will generate a zip file with what would be published for the package - it will show you in the console output all the files that are included

mattkim93 (commented):
neat, very clear.
This is the first instance of storybook I'm seeing in this repo -- super useful 🤔 I might bring that back to the subs team once I'm done here.

LGTM once the .gitignore is added

blocks/identity-block/components/FormInputField/index.story.jsx
nschubach: npmignore is now ncluded

JackHowa (approved):

JackHowa:
looks good

PR #1054, opened 2021-08-19
Remove nonce content source as not required

matthewroach:
Removal of the nonce content source from the identity block as not required. Using a content source as so would cause unwanted fusion caching


JackHowa (approved):
if this is an issue, makes sense

PR #1063, opened 2021-08-24
Fix input field

nschubach:
## Description
Fixing an issue in the form input field and setting up for the confirm password component.

## Test Steps
1. Checkout this branch `git checkout fix-input-field`
2. Run fusion repo with linked blocks `npm run storybook`

### Before
Setting a custom error would cause the error state to get stuck in error.

### After
Setting a custom error is now properly cleared before checking validity.

## Author Checklist

## Reviewer Checklist

### DIFF
--- a/blocks/identity-block/components/FormInputField/index.jsx
+++ b/blocks/identity-block/components/FormInputField/index.jsx
@@ -25,25 +25,27 @@ const FormInputField = ({
   validationPattern,
 }) => {
   const [valid, setValid] = useState(true);
+  const [value, setValue] = useState(defaultValue);
+  const [initialBlur, setInitialBlur] = useState(showDefaultError);
   const inputElement = useRef();
 
   useEffect(() => {
-    if (validationErrorMessage) {
-      inputElement.current.setCustomValidity(validationErrorMessage);
+    if (initialBlur) {
+      inputElement.current.setCustomValidity('');
+      const isValid = inputElement.current?.checkValidity();
+      if (!isValid && validationErrorMessage) {
+        inputElement.current.setCustomValidity(validationErrorMessage);
+      }
+      setValid(isValid);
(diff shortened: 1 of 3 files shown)


JackHowa:
Screen Shot 2021-08-24 at 13 08 54
Screen Shot 2021-08-24 at 13 08 49
looks good

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -25,34 +25,35 @@ const FormInputField = ({
   validationPattern,
 }) => {
   const [valid, setValid] = useState(true);
+  const [value, setValue] = useState(defaultValue);
```
JackHowa, line 28: wait, aren't we tracking state in the parent? this is new

nschubach: yeah, I needed to hold that value to pass it along to the useEffect so that a change will trigger a revalidation.  I really didn't want to hold this value internally, but I don't see a way around it.

blocks/identity-block/components/FormInputField/index.jsx
```diff
@@ -74,6 +75,7 @@ const FormInputField = ({
           ...(!valid ? ['xpmedia-form-field-input--error'] : []),
         ].join(' ')}
         id={inputId}
+        name={name}
```
JackHowa, line 80: was this change deliberate? with field parameters

nschubach: Yeah.  Name is required since it's used in the id for the field group so it HAS to be there.
```
  const infoId = `id_${name}_error`;
  const inputId = `id_${name}_input`;
```

JackHowa (approved):
looks good -- good find

PR #1080, opened 2021-09-10
Add chromatic ignore class to embeds

matthewroach:
Added class `chromatic-ignore` to embeds in article chain to try and overcome false positives in chromatic runs


JackHowa (approved):
good idea

PR #1094, opened 2021-09-20
Update paths to include all .scss files

matthewroach:
Update .scss path in beta-build action file to be the same as other extension to ensure it picks up on .scss file changes


JackHowa (approved):
my bad, must've missed this one

PR #1095, opened 2021-09-20
ci: Fix scss typo for watching files

JackHowa:
- fix watching scss files per #1094


matthewroach:
Not needed as the beta PR will come back to canary #1096

PR #1096, opened 2021-09-20
ci: Fix scss typo for watching files

JackHowa:
- made pr off of beta as well as canary depending on what's easier for ya @matthewroach 
- #1095 canary pr


matthewroach:
Let's do to beta, as will need the stable updates when we push to production

matthewroach (approved):

PR #1150, opened 2021-10-29
POC Setup codespace with npm and storybook 

JackHowa:
- Watching github universe, I set this up with devcontainer startup using secrets for read-only package token and running storybook on start https://www.githubuniverse.com/2021/v/s-707713?i=BspKY7zE38ZbIP24PJlqSxbxs6nkki06
- To test, open it up in codespace! *You may need to create a codespace from here

Screen Shot 2021-10-29 at 11 01 58

- Look at ports for labelled storybook forwarded port

Screen Shot 2021-10-29 at 11 02 56

- See storybook running and reacts to changes

Screen Shot 2021-10-29 at 11 03 35


brentmiller1973 (commented):
I has a chance to look this over the weekend....Awesome!  I would just suggest that if we want to merge this in, that we get this documented as well.

JackHowa:
todo: We will need to update the secrets as well

PR #1164, opened 2021-11-16
Tmedia 558 block code generator

JackHowa:
Goal: 

- Improve themes developer flow of creating a new block, minimizing copy-paste and busywork

todo: 

- [x] Test linking the block 
- [x] Add translation functionality example 
- [x] Get a block property via arc 
- [x] Ingest an from the engine theme sdk. Clients often don't know about this

Test: 

- `npm i`
- `npm run generate:feature`
- Enter the name of your feature based on instructions. Like 'custom promo'
- `npm t` to run its test and others
- `npm run storybook` to see the component in storybook

Screen Shot 2021-11-17 at 17 20 03
Screen Shot 2021-11-17 at 17 19 10


_templates/feature/new/package.ejs.t
```diff
@@ -0,0 +1,29 @@
...
+    "registry": "https://npm.pkg.github.com/",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "ssh://git@github.com/WPMedia/fusion-news-theme-blocks.git",
+    "directory": "blocks/<%= h.inflection.dasherize(block_name) %>-block"
+  },
+  "peerDependencies": {
+    "@arc-fusion/prop-types": "^0.1.5"
```
matthewroach: If we are wanting to have the initial generated code have Engine theme SDK we will need to add it as a peer dependency here

_templates/feature/new/package.ejs.t
```diff
@@ -0,0 +1,29 @@
...
+{
+  "name": "@wpmedia/<%= h.inflection.dasherize(block_name) %>-block",
+  "version": "0.1.0",
+  "description": "<%= h.changeCase.title( block_name ) %> Block",
+  "homepage": "https://github.com/WPMedia/fusion-news-theme-blocks",
+  "license": "CC-BY-NC-ND-4.0",
+  "main": "index.js",
+  "files": [
+    "features",
+    "intl.json"
```
matthewroach, line 14: No `intl.json` is generated with this - should we generate one?

_templates/feature/new/default-entry.ejs.t
```diff
@@ -0,0 +1,42 @@
...
+import React from 'react';
+import PropTypes from '@arc-fusion/prop-types';
+
+// icons and other utilities can be found in the engine theme sdk storybook
+// https://beta--5eed0506faad4f0022fedf95.chromatic.com/
+import {
+  EnvelopeIcon,
+} from '@wpmedia/engine-theme-sdk';
+
+function <%= h.changeCase.pascal(block_name) %>(props) {
```
matthewroach: Can we switch the syntax to be

```
const <%= h.changeCase.pascal(block_name) %> = ({ customFields }) => {
```

this makes it more consistent with how existing blocks are built

matthewroach (changes requested):
Some initial comments added

_templates/feature/new/package.ejs.t
JackHowa: oops, yep good call

_templates/feature/new/package.ejs.t
JackHowa: I think so. I was thinking of showing an example translation

_templates/feature/new/default-entry.ejs.t
JackHowa: since client devs may not be familiar with props, didn't want to do too much destructuring. I also put in an explainer with what props are

matthewroach: This is within our repo - and should act as a "best practise"

_templates/feature/new/package.ejs.t
JackHowa: updated

_templates/feature/new/default-entry.ejs.t
JackHowa: updated

_templates/feature/new/package.ejs.t
JackHowa: ok added one; updated. included your doc on lokalise as well

matthewroach (approved):

PR #1208, opened 2022-01-03
Tmedia 595 - Results List with Collection cannot have more than 10 entries

brentmiller1973:
## Description
When using the Results List Block from Themes with the Collections content-source, you cannot get a size over 10, it will just show nothing instead.
Considering the collections-api can serve up to 20 entries, I don't understand why.
I was able to reproduce it on 2 environments, Sandbox. All OOTB Themes, none of the examples are currently live - but it can be easily reproduced on Themes environment.

## Jira Ticket
[- [TMEDIA-](https://arcpublishing.atlassian.net/browse/TMEDIA-)](https://arcpublishing.atlassian.net/browse/TMEDIA-595)

## Acceptance Criteria
Any size up to 20 displays in the Results List with a Collection.

## Test Steps

1. Checkout this branch `git checkout tmedia-595`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/collections-content-source-block,@wpmedia/resizer-image-block,@wpmedia/results-list-block
`
3. Create a new page or use and existing one and ensure a Results List Block is added.
4. In the configure content section for the result list, add the following properties:
image
5. Save and publish changes and then view the published page under the Gazette profile.

### Before
Any setting over 10 would not show any entries in the result list.

### After
Any setting over 10 is now shown.  However, if the entry is over 20, the number is trimmed to 20 because of the limitation of the collection API being capped at 20 entries.

If the user configures the results list with 20 entries, the show more button is not shown.  I was unable to fix this. One thing I tried is that since an entry of 10-19 was showing the show more button, I tried to cap an entry of 20 to 19, but that did not seem to work (probably due to client hydration taking over?).  

## Author Checklist
- [x ] Confirmed all the test steps a reviewer will follow above are working.
- [x ] Confirmed there are no linter errors. Please run `npm run lint` to check for errors. Often, `npm run lint:fix` will fix those errors and warnings.
- [ x] Ran this code locally and checked that there are not any unintended side effects. For example, that a CSS selector is scoped only to a particular block.
- [x ] Confirmed this PR has reasonable code coverage. You can run `npm run test:coverage` to see your progress.
  - [x ] Confirmed this PR has unit test files
  - [x ] Ran `npm run test`, made sure all tests are passing

## Reviewer Checklist


matthewroach (dismissed):
Fundamentally this fixes the problem at hand, but it doesn't truly fix the issue that's affecting the results list block.

Things we currently know
1. The results list block isn't rendering collection items if you use a size over 10
2. The collections api content source **does** works outside of the result list block if you specify a value over 10 and 20 or less - using a value of 20 via the content source returns an error from the underlying API.
   a. This can be tested via the content source debugger

---

Based on the items we currently know, there is two issues we need to address
1. Should we cap the size of the collection API? It is valid as otherwise the user will see errors. This could be seen as an editor/set up issue - maybe some messaging around limits on collection API calls

   a. For now I think capping the limit is valid

2. The results list block **only** isn't rendering items if a user specifies a valid number between 11 and 20 within the `size` field of the content source - even when this is a valid number

    a. This being a results list block specific issue made me dive into **why** - and the why is based on the data fetching logic implemented within the results list block. The fetching logic implemented is what is causing the underlying issue. The results list block has logic to "prefetch" the next data set be requesting twice as much data as the user asked for by doing `configuredSize * 2` but as the collections API can only return a maximum of 20 results the response is an error - You can see this by looking at the `size` attribute of the content call, if you set the size to 11 the content call is setting the size attribute to 22. By implementing the cap via the content source directly this does "fix" the issue, by the results list block pre fetching mechanics is now different, as the request will always have 20 has the cap is set within the API even though the block is request say 22.
  
    b. The previous points is the reason why we now loose the "see more" functionality when the results list block is set to use collections with a size of 20 as the pre fetch of the next results set never happened as thus the block thinks there is only 20 results

brentmiller1973:
> Fundamentally this fixes the problem at hand, but it doesn't truly fix the issue that's affecting the results list block.
> 
> Things we currently know
> 
> 1. The results list block isn't rendering collection items if you use a size over 10
> 2. The collections api content source **does** works outside of the result list block if you specify a value over 10 and 20 or less - using a value of 20 via the content source returns an error from the underlying API.
>    a. This can be tested via the content source debugger
> 
> Based on the items we currently know, there is two issues we need to address
> 
> 1. Should we cap the size of the collection API? It is valid as otherwise the user will see errors. This could be seen as an editor/set up issue - maybe some messaging around limits on collection API calls
>    a. For now I think capping the limit is valid
> 2. The results list block **only** isn't rendering items if a user specifies a valid number between 11 and 20 within the `size` field of the content source - even when this is a valid number
>    a. This being a results list block specific issue made me dive into **why** - and the why is based on the data fetching logic implemented within the results list block. The fetching logic implemented is what is causing the underlying issue. The results list block has logic to "prefetch" the next data set be requesting twice as much data as the user asked for by doing `configuredSize * 2` but as the collections API can only return a maximum of 20 results the response is an error - You can see this by looking at the `size` attribute of the content call, if you set the size to 11 the content call is setting the size attribute to 22. By implementing the cap via the content source directly this does "fix" the issue, by the results list block pre fetching mechanics is now different, as the request will always have 20 has the cap is set within the API even though the block is request say 22.
>    b. The previous points is the reason why we now loose the "see more" functionality when the results list block is set to use collections with a size of 20 as the pre fetch of the next results set never happened as thus the block thinks there is only 20 results

@matthewroach I tried inititally capping the amount in https://github.com/WPMedia/arc-themes-blocks/blob/b54ca1d1996bec8c49f3632877586a5902343240/blocks/results-list-block/features/results-list/results/index.jsx#L64 Basically, similar to what I did in the API:

`      case 'content-api-collections': {
        let updatedSize = size;
        if (updatedSize && updatedSize > 9) updatedSize = size < 20 ? size : 20;
        return { from: offset, updatedSize };
      }`
However, this gave worse behavior in that the see more button was showing intermittently on any size configuration.

matthewroach:
@visionbegin It's likely there would be more changes needed to the block for the button to show, as the button is shown when it knows there are more results - and as the collection limit is 20 if a user sets the size to 20 the block never fetches the next 20 under the covers, we would need to add logic to account for the collection limit of 20 and pre fetch more results

brentmiller1973:
> @visionbegin It's likely there would be more changes needed to the block for the button to show, as the button is shown when it knows there are more results - and as the collection limit is 20 if a user sets the size to 20 the block never fetches the next 20 under the covers, we would need to add logic to account for the collection limit of 20 and pre fetch more results

@matthewroach I've looked as this already last week....one attempt was to curtail a setting of 20 down to 19,18...15 and that doesn't trigger the logic needed to show the more button.  Any insights would be appreciated.

matthewroach:
> @matthewroach I've looked as this already last week....one attempt was to curtail a setting of 20 down to 19,18...15 and that doesn't trigger the logic needed to show the more button. Any insights would be appreciated.

@visionbegin - I've created a PR showing how we could do fetching the next set of data for results list block via the collections content source by switching to using fetch

https://github.com/WPMedia/arc-themes-blocks/pull/1212

blocks/collections-content-source-block/sources/content-api-collections.js
```diff
@@ -1,26 +1,64 @@
...
+    method: 'GET',
+  }).then(({ data: content }) => {
+    if (getNext === 'false') {
+      return content;
+    }
+
+    const existingData = content;
+
+    return axios({
+      url: `${CONTENT_BASE}content/v4/collections?${_id ? `_id=${_id}` : `content_alias=${contentAlias}`}${site ? `&website=${site}` : ''}&from=${updatedSize}${size ? `&size=${updatedSize}` : ''}&published=true`,
```
ecarlisle: @matthewroach : We also need a trailing slash here.

ecarlisle (changes requested):

blocks/collections-content-source-block/sources/content-api-collections.js
matthewroach: Added

JackHowa:
is this good to merge @ecarlisle @matthewroach ? the other ticket was merged into this one?

ecarlisle (approved):

PR #1216, opened 2022-01-05
Remove unnecessary boolean logic with typeof

JackHowa:
refactor: Removes unnecessary boolean logic. 

typeof will always return a truthy string value https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof


blocks/related-content-content-source-block/sources/related-content.js
```diff
@@ -20,7 +20,7 @@ export default {
     _id: 'text',
   },
   transform: (data, query) => {
-    if (data && data.basic && typeof Array.isArray(data.basic)) {
+    if (data && data.basic) {
```
matthewroach: This is removing the logic of checking that `data.basic` is an array and the code below would error out. `typeof` would always return `true` in the current logic but just removing the `typeof` check and leaving the `Array.isArray()` check would make line act as what it appears that was it's intention

matthewroach (changes requested):

blocks/related-content-content-source-block/sources/related-content.js
JackHowa: ah yeah. looks like that was the intent based on the test but must've been a typo: 

```
    it('should not call getResizedImageData if data is not array', () => {
      const { default: contentSource } = require('./related-content');
      contentSource.transform({ basic: '' }, {});
      expect(mockFn.mock.calls.length).toBe(0);
    });
```

matthewroach (approved):

PR #1210, opened 2022-01-04
TMEDIA-580 test only changed files on push and in actions

JackHowa:
## Description
Test locally only changed files on push* to improve dev productivity

## Jira Ticket
- [TMEDIA-580](https://arcpublishing.atlassian.net/browse/TMEDIA-580)

## Acceptance Criteria
- On developer push locally, testing should only occur on files that changed vs canary base branch
- GitHub Actions run on changed as well rather than all of the tests on prs to canary only

## Test Steps

1. Checkout this branch `git checkout TMEDIA-580-test-commit-changed`
2. Reinstall your deps `rm -rf node_modules && npm i` for new jest version
2. In one file, like byline, break the tests: `const ArticleByline = () => <Byline font="Primary" />;` -> `const ArticleByline = () => <p>by</p>;`
3. stage the file `git add blocks/byline-block/features/byline/default.jsx`
4. Commit the file `git commit -m "only byline" `
5. Break the tests of another file
6.  Add the file and commit the file
7. Try to `git push`. See the linting run (deferred to future ticket). See the only files changed are being tested. This should fail
8. To reset your branch to mine, run `git reset --hard origin/TMEDIA-580-test-commit-changed`

### Before

- On push, every test is run, taking minutes 
- Github actions test take a long time 

### After

- On push, only relevant files are tested, taking seconds (depending on how many files changed)
- Github action testing speed is improved as well
- See github action running and failing with two commits on files https://github.com/WPMedia/arc-themes-blocks/runs/4707169538?check_suite_focus=true
- See passing github action https://github.com/WPMedia/arc-themes-blocks/runs/4731349155?check_suite_focus=true

## Dependencies or Side Effects
- I removed the output type-specific coverage as new code there will require high code coverage anyway. Also, I didn't want to diverge collect coverage didn't seem to work as expected when there were no files changed https://github.com/facebook/jest/issues/5580 https://github.com/facebook/jest/issues/8474
- Linting is still on push all files. This will be addressed in the future
- Had to update jest due to a bug with onlyChanged in the GitHub action https://github.com/facebook/jest/issues/9602
- Only targets the canary branch for changes and tests

## Author Checklist
_The author of the PR should fill out the following sections to ensure this PR is ready for review._
- [x] Confirmed all the test steps a reviewer will follow above are working.
- [x] Confirmed there are no linter errors. Please run `npm run lint` to check for errors. Often, `npm run lint:fix` will fix those errors and warnings. -> na
[...]

PR #1218, opened 2022-01-06
POC for working test Tmedia 580 test working GitHub changeset (testing working github action)

JackHowa:
- Test that github action into canary works as expected


blocks/card-list-block/features/card-list/default.jsx
```diff
@@ -179,7 +181,7 @@ const CardListItems = (props) => {
               </article>
               {contentItems.slice(1).map((element) => {
                 const {
-                  headlines: { basic: headlineText } = {},
+                  headlines: { basic: headlineText },
```
JackHowa, line 184: removed some of these fallbacks just bc they weren't tested and wanted to get a positive coverage quick

matthewroach: Removing the `= {}` will mean that the headlines key is always available with basic key

JackHowa: this is to just test a positive change in the github action

JackHowa: see note in https://github.com/WPMedia/arc-themes-blocks/pull/1210 See passing github action https://github.com/WPMedia/arc-themes-blocks/runs/4731349155?check_suite_focus=true

PR #1238, opened 2022-01-25
POC perf: Show placeholder area if no headline or not visible

JackHowa:
- See resulting ticket from investigation https://arcpublishing.atlassian.net/browse/TMEDIA-612


JackHowa:
Not the solution we want to go for with always keeping this space

PR #1241, opened 2022-01-26
Chromatic - Header Nav Chain - Delay for logo

matthewroach:
Trying out using the chromatic delay parameter to get some consistency with logo loading in header chain story


JackHowa (approved):
Looks good, nice work. Even visually it looks like there's a bit of delay so this makes sense. 

I don't recall there being delay when I first set this is up so this is interesting; hope this will be a solution

PR #1277, opened 2022-02-18
Add 768 chromatic viewport for Top Table list

matthewroach:
There was a report of an issue with the top table list medium promo rendering incorrectly at 768px - adding a story to have a baseline of what the top table list looks like today at that viewport


JackHowa (approved):
Good idea

PR #1298, opened 2022-03-21
Tmedia 701 raw html static

JackHowa:
## Description

Use static for lead art raw html

- NOTE: Article body chain already uses static so this work was deferred for using raw_html https://github.com/WPMedia/arc-themes-blocks/blob/arc-themes-release-version-1.20/blocks/article-body-block/chains/article-body/default.jsx#L204 https://github.com/WPMedia/arc-themes-blocks/blob/arc-themes-release-version-1.20/blocks/article-body-block/chains/article-body/_children/html.jsx#L13

## Jira Ticket

- [TMEDIA-701](https://arcpublishing.atlassian.net/browse/TMEDIA-701)

## Acceptance Criteria

When a component is outputted on a page, fusion may re-render it a few times as state changes. For an HTML block, we don't want that and want React to ignore it.

## Test Steps

1. Checkout this branch `git checkout TMEDIA-701-raw-html-static`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/lead-art-block`
3. Pull down the story with raw html and a script tag in the lead art https://corecomponents.arcpublishing.com/composer/edit/VX2JWE3AFRBXXIR74Z3UICUGBU/
4. Create a page with the lead art block within an article body chain
5. Set the content source to be content-api with the id `VX2JWE3AFRBXXIR74Z3UICUGBU` 
6. Publish the page 
7. See the fusion static tags around the content. A unique id to the client- and server-side is also present
Screen Shot 2022-03-22 at 15 53 42
8. Validate in storybook that the html still renders as expected under lead art
Screen Shot 2022-03-22 at 15 58 51

# Questions Answered

- Is putting a script tag around console.log the best way to test this? may need to cause a state change like with the image lightbox open and close to ensure this is properly not re-rendering. -> eh it's complicated. ensure that the fusion tags for static surround the content
- it looks like there's a mainContent that goes into the Lightbox in engine theme sdk. Then there's also a dangerously set html there as well? Might need to investigate more how that works.  -> not used in raw html
- I'm assuming then that this will be merged in 1.2 as well? -> yes

### Before

The html within was not using fusion static and could re-render 

### After

Based on using static with engine, the content inside the static tag will not re-render

- Removed unreachable code with the lightbox and raw html

## Author Checklist

  - [ ] Confirmed this PR has unit test files
  - [ ] Ran `npm run test`, made sure all tests are passing

## Reviewer Checklist

### DIFF
--- a/blocks/lead-art-block/features/leadart/default.jsx
+++ b/blocks/lead-art-block/features/leadart/default.jsx
@@ -141,6 +141,12 @@ class LeadArt extends Component {
 						customFields={{
 							playthrough: !!customFields?.playthrough,
 						}}
+						displayTitle={!hideTitle}
+						displayCaption={!hideCaption}
+						displayCredits={!hideCredits}
+						subtitle={lead_art?.headlines?.basic}
+						caption={lead_art?.description?.basic}
+						credits={lead_art.credits}
 					/>
 				);
 			}
@@ -260,19 +266,22 @@ LeadArt.propTypes = {
 		}),
 		...videoPlayerCustomFields(),
 		hideTitle: PropTypes.bool.tag({
-			description: "This display option applies to Lead Art media types: Images and Gallery",
+			description:
+				"This display option applies to Lead Art media types: Images, Gallery, and Video",
 			label: "Hide Title",
 			defaultValue: false,
 			group: "Display Options",
 		}),
 		hideCaption: PropTypes.bool.tag({
-			description: "This display option applies to Lead Art media types: Images and Gallery",
+			description:
+				"This display option applies to Lead Art media types: Images, Gallery, and Video",
 			label: "Hide Caption",
 			defaultValue: false,
 			group: "Display Options",
 		}),
 		hideCredits: PropTypes.bool.tag({
-			description: "This display option applies to Lead Art media types: Images and Gallery",
+			description:
+				"This display option applies to Lead Art media types: Images, Gallery, and Video",
 			label: "Hide Credits",
 			defaultValue: false,
 			group: "Display Options",
--- a/blocks/lead-art-block/features/leadart/default.jsx
+++ b/blocks/lead-art-block/features/leadart/default.jsx
@@ -99,24 +99,6 @@ class LeadArt extends Component {
 			let caption = null;
 
 			if (lead_art.type === "raw_html") {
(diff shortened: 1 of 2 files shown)


blocks/lead-art-block/features/leadart/default.jsx
```diff
@@ -135,12 +141,6 @@ class LeadArt extends Component {
 						customFields={{
 							playthrough: !!customFields?.playthrough,
 						}}
-						displayTitle={!hideTitle}
```
matthewroach: Why are these being removed?

matthewroach:
@JackHowa For your questions around the lightbox, from looking at the code this appears to be redundant code and is not reachable when a consumer is using `raw_html` with the lead art

blocks/lead-art-block/features/leadart/default.jsx
JackHowa: This was an accident mid-wip. And it's been fixed for the files affected

JackHowa:
> For your questions around the lightbox, from looking at the code this appears to be redundant code and is not reachable when a consumer is using raw_html with the lead art

yeah I agree. Looks like we were in the process of removing this as well previously. happy to clean it up fully https://github.com/WPMedia/arc-themes-blocks/commit/8cb580e104377f09457653112f774dac031c353b 

```
				// this could be figure and figcaption, a react component
				const mainContent = (
					<>
						<div dangerouslySetInnerHTML={{ __html: lead_art.content }} />
					</>
				);
				lightbox = (
					<>
						{isOpen && (
							<Lightbox
								mainSrc={mainContent}
								onCloseRequest={this.setIsOpenToFalse}
								showImageCaption={false}
							/>
						)}
					</>
				);
```

JackHowa:
```
Is putting a script tag around console.log the best way to test this? 
```

Still not sure if this is the best way to test this. After all, we're dangerously-setting the html, which feels like it should execute client-side.  @matthewroach

JackHowa:
> ```
> Is putting a script tag around console.log the best way to test this? 
> ```
> 
> Still not sure if this is the best way to test this. After all, we're dangerously-setting the html, which feels like it should execute client-side. @matthewroach

Based on feedback, validating that the fusion tag is surrounding the content is enough.

matthewroach (approved):

PR #1315, opened 2022-04-14
Tbrands 28 carousel block

brentmiller1973:
## Issues and observations
1. The stylint in this repo doesn't like the BEM naming convention.
1. We will need translations for "step # of #"

## Big Issues

_Note if you want to run this, use this branch of the theme pack:
https://github.com/WPMedia/arc-themes-feature-pack/pull/220
 and start with `npx fusion start -f -l @wpmedia/story-carousel-block,@wpmedia/single-column-layout-block` (there is a fix in single-column-layout-block)_

1. **Arrow positioning:** We need to be able to control the position of the arrow bar (`.carousel_actions`) vertically.  In the design, it's not always vertically centered:

image

2. **Responsive:** The carousel component provides a prop called `slidesToShow`.  This seems to suggest that the carousel takes a desktop-first approach to responsiveness, however on mobile it does not collapse to one or one and  half slides at a time. 
 To handle mobile, I provided props on the block level...`desktopSlideNumber` and `mobileSlideNumber `. Dependending on the breakpoint, these values will get set on `slidesToShow` that is passed into the carousel component.  To determine the breakpoint: First a new breakpoint token was added to the themefile in the themepack: https://github.com/WPMedia/arc-themes-feature-pack/blob/c1d626b48310a8b5b72af156ecc73585fe215887/site-styles/arc-commerce/_index.scss#L8. Second a useEffect is used to read in the prop, call a function that  checks the window size against the breakpoint as well as set up a window listener that will call that same function when the window resizes.  That useEffect can be found here:  https://github.com/WPMedia/arc-themes-blocks/blob/f8f075540608f15a01c63457c7efd3ec79c00c38/blocks/story-carousel-block/features/story-carousel/default.jsx#L40. This solution does transition well from desktop to mobile and gets us closer to the mobile design:

image

3) **Arrow logic and excessive scrolling.** See the video below for a demonstration and explanation for desktop.  The video after that will illustrate the issues with mobile that are slightly different:

https://user-images.githubusercontent.com/833461/163680067-ccbe75b6-ad6f-4569-9b28-716b64fb13b2.mov

https://user-images.githubusercontent.com/833461/163680130-25783363-f6de-47eb-bdcb-f024901dda5e.mov

## Jira Ticket

[- [TMEDIA-](https://arcpublishing.atlassian.net/browse/TMEDIA-)](https://arcpublishing.atlassian.net/browse/TBRANDS-28)


blocks/story-carousel-block/features/story-carousel/default.jsx
```diff
@@ -0,0 +1,138 @@
...
+		rightArrowLabel,
+		leftArrowLabel,
+	},
+}) => {
+	// const { locale = "en" } = getProperties(arcSite);
+	// const phrases = getTranslatedPhrases(locale);
+	const content = useContent({
+		source: contentService,
+		query: contentConfigValues,
+	});
```
matthewroach, line 37: PageBuilder is giving errors around `content_elements` this is because later in the block you are destructing the object but in some cases there is no object - we should set an OR to a blank object to stop further code erroring

```suggestion
	}) || {};
```

blocks/story-carousel-block/features/story-carousel/default.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+	// const buildCarousel = () => {
+	//
+	// }
+	return (
+		<div className={BLOCK_CLASS_NAME}>
+			{headline && headline.trim() !== "" && <h1>{headline}</h1>}
+
+			{elements && elements.length && (
```
matthewroach, line 108: Need to use a ternary otherwise you'll see a `0` output in some cases

```suggestion
			{elements && elements.length ? (
```

blocks/story-carousel-block/features/story-carousel/default.jsx
```diff
@@ -0,0 +1,138 @@
...
+// import getTranslatedPhrases from "fusion:intl";
+const BLOCK_CLASS_NAME = "b-story-carousel";
+
+const StoryCarousel = ({
+	arcSite,
+	customFields: {
+		carouselContentConfig: { contentService = "", contentConfigValues = {} } = {},
+		headline,
+		carouselLabel = "Carousel of Images",
+		identifier,
```
matthewroach, line 20: For the identifier we can switch this from a custom field and use the id returned in the `useComponentContext()` hook

```
import { useComponentContext } from "fusion:context";

```

blocks/story-carousel-block/features/story-carousel/default.jsx
```diff
@@ -0,0 +1,138 @@
...
+					}
+					previousButton={
+						<Carousel.Button id={identifier} label={leftArrowLabel}>
+							<Icon name="ArrowLeft" />
+						</Carousel.Button>
+					}
+				>
+					{elements.map((item, index) => {
+						const currentSlide = index + 1;
+						// Do I need to set viewable????????????????
```
matthewroach: Not needed

blocks/story-carousel-block/features/story-carousel/default.jsx
```diff
@@ -0,0 +1,217 @@
...
+	const getParentClassNames = () => {
+		if (elements && elements.length > 4) {
+			return `${BLOCK_CLASS_NAME} ${BLOCK_CLASS_NAME}__story-card-5-plus`;
+		}
+		return `${BLOCK_CLASS_NAME} ${BLOCK_CLASS_NAME}__story-card-4-less`;
+	};
+
+	return (
+		<Stack className={getParentClassNames()}>
+			{headline && headline.trim() !== "" && <Heading>{headline}</Heading>}
```
matthewroach, line 106: We will need to wrap the block in a `<HeadingSection>` otherwise you'll have a h1 output here

```suggestion
			{headline && headline.trim() !== "" ? <Heading>{headline}</Heading> : null}
```

blocks/story-carousel-block/features/story-carousel/default.jsx
```diff
@@ -0,0 +1,217 @@
...
+					}
+				>
+					{elements.map((item, index) => {
+						const currentSlide = index + 1;
+						return (
+							<Carousel.Item key={item._id} label={`Slide ${currentSlide} of ${elements.length}`}>
+								<a className={`${BLOCK_CLASS_NAME}__story-card`} href={item.canonical_url}>
+									<Image alt={item.headlines.basic} src={testImage} />
+									{item.headlines?.basic ? (
+										<Heading
```
matthewroach, line 131: You'll need to add a `<HeadingSection>` wrapper to ensure the heading here is incremented correctly

PR #1323, opened 2022-04-29
TMEDIA-847 - Add new spacing token

matthewroach:
## Description

Add a new spacing token and bump all existing values to realign with the original intentions

## Jira Ticket

- [TMEDIA-847](https://arcpublishing.atlassian.net/browse/TMEDIA-847)


JackHowa (approved):
Makes sense that it's more linear

JackHowa (approved):
More linear. Looks good to add another

PR #1362, opened 2022-05-27
Tmedia 732 fix import tests side effects

JackHowa:
How to test: 
- ` npm t` 
- All tests should pass

Fixes https://github.com/WPMedia/arc-themes-blocks/runs/6627371706?check_suite_focus=true

### DIFF
--- a/blocks/byline-block/README.md
+++ b/blocks/byline-block/README.md
@@ -60,8 +60,6 @@ This block does not emit any events.
 
 ## Additional Considerations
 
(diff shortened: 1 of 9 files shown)


nschubach (approved):
All tests seem to run fine now.  Inappropriate mocks are interesting though.

These blocks also inappropriately have peer dependencies to the byline (which they also do not use).  I assume that's not impacting anything, but they should be removed:
image

JackHowa:
@nschubach updated to remove peer dep reference

PR #1365, opened 2022-05-31
TMEDIA-736 : Share Bar Block

ecarlisle:
## Description

1. Share bar has been updated to new architecture.
2. Testing has been updated from Mocha to React Testing Library (100% coverage).
3. Storybook token files has been updated with block styles.
4. Feature Pack update PR can be found at https://github.com/WPMedia/arc-themes-feature-pack/pull/263 .

## Jira Ticket

- [TMEDIA-736 : Share Bar block](https://arcpublishing.atlassian.net/browse/TMEDIA-736)

## Acceptance Criteria

No criteria listed on ticket. See Description section.

## Test Steps

1. Checkout this branch `git checkout TMEDIA-736-share-bar-block`
2. Start Storybook with `npm run storybook`.
3. Verify both functionality and cosmetics of Share Bar block. Block will not display on mobile but will display on desktop.
image
5. Verify test coverage of branch as well as lint compliance. 

## Effect Of Changes

Block has been updated to new architecture and styling.

## Author Checklist

  - [X] Ran `npm run test`, made sure all tests for this branch are passing
- [X] Confirmed relevant documentation has been updated/added.

## Reviewer Checklist


blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -135,7 +117,17 @@ export const ShareBar = ({
 		return null;
 	}
 
-	return <div className="ts-share-bar">{shareButtons}</div>;
+	return (
+		<Stack
+			className={BLOCK_CLASS_NAME}
+			alignment="center"
+			direction="vertical"
+			gap="0"
```
matthewroach: No needed

.storybook/themes/news.scss
```diff
@@ -835,6 +844,62 @@
...
+					width: 44px,
+					components: (
+						button-medium: (
+							background: transparent,
+							border-width: 0,
+							margin: 5px,
+							padding: 0,
+						),
+						icon: (
+							height: var(--global-font-size-9),
```
matthewroach: Should be using `spacing` tokens not font size tokens

.storybook/themes/news.scss
```diff
@@ -835,6 +844,62 @@
...
+					left: 0,
+					padding: 10px 0,
+					position: fixed,
+					top: 200px,
+					width: 44px,
+					components: (
+						button-medium: (
+							background: transparent,
+							border-width: 0,
+							margin: 5px,
```
matthewroach: We should not be setting margin on the button - we should make use of the `Stack` `gap` property to add spacing between the buttons

.storybook/themes/news.scss
```diff
@@ -835,6 +844,62 @@
 					padding: 0 var(--global-spacing-6) 0 0,
 					display: flex,
 				),
+				share-bar: (
+					border-radius: 0 5px 5px 0,
```
matthewroach: Can we use the `border-radius` token here?

.storybook/themes/news.scss
```diff
@@ -633,6 +633,15 @@
 				right-rail-advanced-main-right-rail-bottom: (
 					grid-area: main-right-rail-bottom,
 				),
+				share-bar: (
```
matthewroach, line 689: Why don't we set all the styles at the mobile breakpoint and then just change the `display` value at the desktop? That way we will not need to specify all `initial` values

.storybook/themes/news.scss
```diff
@@ -835,6 +844,62 @@
...
+						),
+					),
+				),
+				share-bar-facebook: (
+					components: (
+						icon: (
+							fill: #4267b2,
+						),
+					),
+				),
```
matthewroach: If you apply these styles at the mobile breakpoint you can write this as 

```suggestion
				share-bar-facebook: (
					fill: #4267b2,
				),
```

blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -99,7 +83,7 @@ const ShareBarContainer = () => {
 };
 
 export const ShareBar = ({
-	customFields = {},
+	customFields,
```
matthewroach: Removing the default ` = {} ` breaks the code as on line 94 the `Object.keys()` is expecting an object which in some cases `customFields` will never be an object

blocks/share-bar-block/features/share-bar/default.test.jsx
```diff
@@ -29,19 +32,30 @@ describe("When the share bar is shown", () => {
 				phrases={mockPhrases}
 			/>
 		);
-		expect(wrapper.find(".ts-share-bar__button")).toHaveLength(5);
+
+		const buttons = container.querySelectorAll(".b-share-bar button");
```
matthewroach: We shouldn't be querying by selectors with RTL (React Testing Library) - we should use `getByRole` to query for specific buttons - see https://testing-library.com/docs/queries/about#priority - for details

blocks/share-bar-block/features/share-bar/default.test.jsx
```diff
@@ -168,9 +208,89 @@ describe("When the share bar is shown", () => {
...
+					bubbles: true,
+					cancelable: true,
+				})
+			);
+			await expect(window.open).toBeCalled();
 			expect(window.location.origin).toEqual("http://localhost");
-			expect(window.open).toBeCalled();
+		});
+
+		/*
```
matthewroach: Is this needed? If not, we can delete

matthewroach (changes requested):

blocks/share-bar-block/package.json
```diff
@@ -19,13 +20,8 @@
...
     "directory": "blocks/share-bar-block"
   },
-  "scripts": {
-    "test": "echo \"Error: run tests from root\" && exit 1",
-    "lint": "eslint --ext js --ext jsx features"
-  },
   "peerDependencies": {
-    "@wpmedia/engine-theme-sdk": "*",
-    "@wpmedia/news-theme-css": "*"
+    "@wpmedia/arc-themes-components": "*"
```
JackHowa, line 25: small note: We should also have arc proptypes as a peer dependency here. Good work documenting what you have so far with arc themes components

JackHowa:
@ecarlisle good work using storybook. I noticed something I hadn't seen before: that the logos are different. I assume that this is intentional. I think this would be good to bring up in the discrepancies doc in case there's questions in the future 

Screen Shot 2022-06-01 at 13 39 16

https://www.chromatic.com/test?appId=5f91e4721ccaba0022a10782&id=6297927f7ed97a004af41aaf

blocks/share-bar-block/features/share-bar/default.jsx
ecarlisle: Thanks! Removed.

.storybook/themes/news.scss
ecarlisle: Thanks! Changed.

.storybook/themes/news.scss
ecarlisle: Sorry, did you mean the `box-shadow` property? If so, I was able to move it over. I had it in the block Sass initially because the value had commas and that was messing up the Sass map in `news.scss`. I had read, though, where you can escape the commas by putting the value in parenthesis. 

cc @visionbegin as we were wondering about that.

ecarlisle: However, if the question was if we can use the `border-radius` prop & value there, it does seem to work. I can't think of why not.

.storybook/themes/news.scss
ecarlisle: I took the margin out and instead used the biggest padding I could for the buttons to make them the biggest tap target I could. By doing that, I didn't need to add a `Stack` `gap`.

.storybook/themes/news.scss
matthewroach: No, I meant the border-radius token. We should not be setting

```
border-radius: 0 5px 5px 0,
```

```
border-radius: 0 var(--border-radius) var(--border-radius) 0
```

blocks/share-bar-block/_index.scss
```diff
@@ -0,0 +1,32 @@
+@use "@wpmedia/arc-themes-components/scss";
+
+.b-share-bar {
+	box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
```
matthewroach: We can remove this if we are setting it via the tokens now

blocks/share-bar-block/_index.scss
```diff
@@ -0,0 +1,32 @@
...
+	@include scss.block-properties("share-bar-pinterest");
+	@include scss.block-components("share-bar-pinterest");
+}
+
+.b-share-bar__twitter {
+	@include scss.block-properties("share-bar-twitter");
+	@include scss.block-components("share-bar-twitter");
+}
+
+.b-share-bar__email {
```
matthewroach, line 28: There is a miss match with the code and the tokens - the code has the following class `b-share-bar__envelope`

blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -60,30 +48,26 @@ const share = {
...
-		case "email":
-			return <EnvelopeIcon fill="#C72A22" />;
-	}
-}
+const getLogoComponent = {
+	facebook: <Icon name="Facebook" className={`${BLOCK_CLASS_NAME}__facebook`} />,
+	linkedIn: <Icon name="LinkedIn" className={`${BLOCK_CLASS_NAME}__linkedin`} />,
+	pinterest: <Icon name="Pinterest" className={`${BLOCK_CLASS_NAME}__pinterest`} />,
+	twitter: <Icon name="Twitter" className={`${BLOCK_CLASS_NAME}__twitter`} />,
+	email: <Icon name="Envelope" className={`${BLOCK_CLASS_NAME}__envelope`} />,
```
matthewroach:
```suggestion
	email: <Icon name="Envelope" className={`${BLOCK_CLASS_NAME}__email`} />,
```

blocks/share-bar-block/features/share-bar/default.test.jsx
ecarlisle: **Outdated. Here for context but please disregard. Follow-up comment below.**

@matthewroach ,
Thanks, great callout! I'm looking at the `aria-label` on each one of the buttons and it doesn't look like right phrase is getting placed in. Are we handling the phrases in `intl.json` differently on this version. Here's what I'm seeing so far:
image

blocks/share-bar-block/_index.scss
ecarlisle: Thanks! Got it! 👍🏼 
I was trying to figure out why the fill wasn't working on this. This fixed it!

.storybook/themes/news.scss
ecarlisle: Good catch! Using `global-spacing-5` now. Thanks!

.storybook/themes/news.scss
ecarlisle: @matthewroach : Thanks, understood! Looking at the global border radiuses we have so far.
https://github.com/WPMedia/arc-themes-feature-pack/blob/news-2.00/site-styles/daily-intelligencer/_index.scss#L128-L130
We have `"border-radius-1": 0.25rem` which is very close (4px instead of 5px), but I was going for parity. @JackHowa showed me the Blocks Design Discrepancies page in Confluence. 
https://arcpublishing.atlassian.net/wiki/spaces/TI/pages/3138945094/Blocks+Design+Discrepancies
I can list both the `border-radius` and `box-shadow` there, as they aren't supported by existing tokens.
Does that sound ok for in intermediate solution?

.storybook/themes/news.scss
ecarlisle: Thanks, got it!

blocks/share-bar-block/features/share-bar/default.jsx
ecarlisle: @matthewroach : Thanks for confirming that! I put it back in added this test to ensure it was getting test coverage:
https://github.com/WPMedia/arc-themes-blocks/blob/TMEDIA-736-share-bar-block/blocks/share-bar-block/features/share-bar/default.test.jsx#L82-L94

blocks/share-bar-block/features/share-bar/default.test.jsx
ecarlisle: @matthewroach : Thanks! That was very helpful to know! There unfortunately wasn't anything visible to the user that identified the buttons except for the SVGs. I took out all the `querySelector` queries and replaced with the following:
1. All buttons selected: https://github.com/WPMedia/arc-themes-blocks/blob/TMEDIA-736-share-bar-block/blocks/share-bar-block/features/share-bar/default.test.jsx#L35-L36
2. Some buttons selected: https://github.com/WPMedia/arc-themes-blocks/blob/TMEDIA-736-share-bar-block/blocks/share-bar-block/features/share-bar/default.test.jsx#L57-L58

blocks/share-bar-block/features/share-bar/default.test.jsx
ecarlisle: It did not affect test coverage. I removed all of these and kept the `expect(window.location.origin).toEqual("http://localhost");`.
Thanks!

blocks/share-bar-block/package.json
ecarlisle: Thanks! Added.

blocks/share-bar-block/_index.scss
ecarlisle: @matthewroach : We don't currently have a token for this box shadow but I will this to the Design Discrepancies document in Confluence.

blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -60,30 +48,26 @@ const share = {
...
+};
 
 const ShareBarContainer = () => {
 	const {
 		customFields,
-		globalContent: { headlines: { basic: headlineString = "" }, website_url: websiteUrl = "" } = {},
+		globalContent: {
+			headlines: { basic: headlineString = "" },
+			website_url: websiteUrl = "",
+		},
```
JackHowa: was the fallback `= {}` removed on purpose?

JackHowa:
```suggestion
		} = {},
```

blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -111,22 +95,19 @@ export const ShareBar = ({
 		if (customFields[social]) {
 			const encodedTitle = encodeURIComponent(headlineString);
 			const encodedUrl = encodeSocialUrl(websiteDomain, websiteUrl, encodedTitle, social);
-
 			const socialType = phrases.t(`global.social-${social.toLowerCase()}`);
 
 			shareButtons.push(
-				<button
+				<Button
+					className={`${BLOCK_CLASS_NAME}__${social.toLowerCase()}`}
```
JackHowa: I'm not sure if this classname will work as expected. it looks like this would interfere with: 

```
const getLogoComponent = {
	facebook: <Icon name="Facebook" className={`${BLOCK_CLASS_NAME}__facebook`} />,
	linkedIn: <Icon name="LinkedIn" className={`${BLOCK_CLASS_NAME}__linkedin`} />,
	pinterest: <Icon name="Pinterest" className={`${BLOCK_CLASS_NAME}__pinterest`} />,
	twitter: <Icon name="Twitter" className={`${BLOCK_CLASS_NAME}__twitter`} />,
	email: <Icon name="Envelope" className={`${BLOCK_CLASS_NAME}__email`} />,
};

Comment (JackHowa) [reply blocks/share-bar-block/features/share-bar/default.jsx]:
```suggestion
```

blocks/share-bar-block/features/share-bar/default.jsx
ecarlisle: @JackHowa : Yes, but it was correct to restore. Thank you for the suggested commit. I've included that change and restored the `={}` fallback.

.storybook/themes/news.scss
JackHowa: @ecarlisle it sounds like @matthewroach is suggesting using the token called `--border-radius`. can you update to use that `Can we use the border-radius token here?`

.storybook/themes/news.scss
```diff
@@ -633,6 +633,46 @@
 				right-rail-advanced-main-right-rail-bottom: (
 					grid-area: main-right-rail-bottom,
 				),
+				share-bar: (
+					box-shadow: (
+						0 4px 8px 0 rgba(0, 0, 0, 0.2),
+						0 6px 20px 0 rgba(0, 0, 0, 0.19),
+					),
+					border-radius: 0 5px 5px 0,
```
JackHowa: sounds like this should use the `--border-radius` token

.storybook/themes/news.scss
```diff
@@ -835,6 +875,9 @@
 					padding: 0 var(--global-spacing-6) 0 0,
 					display: flex,
 				),
+				share-bar: (
+					display: flex,
```
JackHowa, line 928: this is redundant right? you're using the stack here so that has display flex. it can also be removed in the mobile version. I'm not sure if I'm missing something

.storybook/themes/news.scss
```diff
@@ -835,6 +875,9 @@
...
+				),
```
JackHowa, line 929:
```suggestion
```

.storybook/themes/news.scss
```diff
@@ -633,6 +633,46 @@
 				right-rail-advanced-main-right-rail-bottom: (
 					grid-area: main-right-rail-bottom,
 				),
+				share-bar: (
+					box-shadow: (
+						0 4px 8px 0 rgba(0, 0, 0, 0.2),
+						0 6px 20px 0 rgba(0, 0, 0, 0.19),
+					),
+					border-radius: 0 5px 5px 0,
+					display: flex,
```
JackHowa:
```suggestion
```

JackHowa: sounds like at one breakpoint there should be display none

.storybook/themes/news.scss
```diff
@@ -633,6 +633,46 @@
...
+					left: 0,
+					padding: 9px 0,
```
JackHowa: padding top and bottom should be spacing 4 based on the mocks 
Screen Shot 2022-06-02 at 15 53 51

.storybook/themes/news.scss
```diff
@@ -633,6 +633,46 @@
 				right-rail-advanced-main-right-rail-bottom: (
 					grid-area: main-right-rail-bottom,
 				),
+				share-bar: (
+					box-shadow: (
```
JackHowa: looks like in the mocks in figma it's using a box-shadow-1 alias 

		"box-shadow-1": 0px 0px 16px rgba(25, 25, 25, 0.4),

Screen Shot 2022-06-02 at 15 55 56

.storybook/themes/news.scss
```diff
@@ -633,6 +633,46 @@
...
+					),
+					border-radius: 0 5px 5px 0,
+					display: flex,
+					left: 0,
+					padding: 9px 0,
+					position: fixed,
+					top: 200px,
+					width: 44px,
+					components: (
+						button-medium: (
```
JackHowa, line 700: looks like we're overriding the inherent styles of the button. looks like this should be a link. think you'll have an easier time using padding and centering it as well with a link

.storybook/themes/news.scss
```diff
@@ -633,6 +633,46 @@
...
+							background: transparent,
+							border-width: 0,
+							padding: 7px 10px,
+							text-align: center,
+						),
+						icon: (
+							height: var(--global-spacing-5),
```
JackHowa: height shouldn't be set in styles. it should be set with props for the theme component

.storybook/themes/news.scss
JackHowa, line 649: We want to be using a link here. it looks like we're overriding the inherent buttons styles anyway. and since this is redirecting we'd want to use an a tag anyway. think you'll have an easier time with padding, etc, eg: 


					components: (
						link: (
							padding: var(--global-spacing-2),
						),
						icon: (
							fill: var(--color-primary),
						),
					),

JackHowa (dismissed):
Looks like there was some design feedback and implementation feedback

.storybook/themes/news.scss
ecarlisle: @JackHowa :  I gave that a try and received the following lint notification:
image
The following URL was provided for context:
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/287854abd066704e2a9964da597e7ab7f6f7e2ad/docs/rules/anchor-is-valid.md

.storybook/themes/news.scss
ecarlisle: @JackHowa : Where can I find these specifications in Figma? Thanks!

.storybook/themes/news.scss
JackHowa: huh oh, didn't realize there was an onclick handler. thought it was a normal redirect with an href

.storybook/themes/news.scss
JackHowa: the variable is also available in the style sheet

.storybook/themes/news.scss
ecarlisle: For this reason I kept the `<Button>` instead of a replacing with a `<Link/>`.

.storybook/themes/news.scss
ecarlisle: @JackHowa : Yes, good catch! I implemented all styles in `default()` and set `display: none`. In `desktop()` I set `display: flex` as the top-level container is a `<Stack/>`.

.storybook/themes/news.scss
ecarlisle: @JackHowa : As said above, this is to set display the `<Stack/>` container from `none` on `default()` to `flex` on `desktop()`.

.storybook/themes/news.scss
ecarlisle: @JackHowa : Got it! I think I was OoO when these global tokens were added and the Figma specs were updated. I'm using `--border-radius` now. Thanks!
image

.storybook/themes/news.scss
ecarlisle: @JackHowa : Thanks! All caught up & using all of these tokens now.

.storybook/themes/news.scss
ecarlisle: @JackHowa : Thanks! All caught up & using all of these tokens now.

.storybook/themes/news.scss
ecarlisle: @JackHowa : Thank you for the commit suggestion. I did change the `display` to `none` for `default()` in my changes.

.storybook/themes/news.scss
ecarlisle: @JackHowa : Now using that token. Thanks!

.storybook/themes/news.scss
ecarlisle: @matthewroach : After further inspection I'm using the `<Stack/>` gap to match the Figma specifications.
https://www.figma.com/file/vWVGgZfZVBdSX7ACo3kmjr/Enhanced-Styling-Design-System?node-id=3507%3A34537

blocks/share-bar-block/README.md
```diff
@@ -1,10 +1,19 @@
 # `@wpmedia/share-bar-block`
 
-_This is the share bar block for the news theme, written as a functional component. Users can choose out of five options (email, facebook, pinterest, linkedin, twitter) which selection of share buttons will be displayed on articles. Clicking on each button will open up a new window._
+This is the share bar block for the news theme, written as a functional component. Users can choose from five options to display on the share bar:
```
matthewroach:
```suggestion
This is the share bar block for displaying icons to allow the page to be shared, written as a functional component. Users can choose from five options to display on the share bar:
```

blocks/share-bar-block/README.md
```diff
@@ -25,26 +34,21 @@ Outline any schema information requirements necessary to know for ths block
...
-| ---------- | --------------------- |
-| `key`      | `english translation` |
+| Phrase key              | Default (English)                           |
+| ----------------------- | ------------------------------------------- |
+| share-button-aria-label | "Share current article via `%{socialType}`" |
+
+**Note:** The `%{socialType}` value is taken from Internationalization fields in the
+[global-phrases-block](https://github.com/WPMedia/arc-themes-blocks/tree/arc-themes-release-version-2.0.1/blocks/global-phrases-block).
 
 ## Events
```
matthewroach: No need to document Events and Event Listening - this block does not emit any events. This documentation is for when a block is emitting custom events

blocks/share-bar-block/_index.scss
```diff
@@ -0,0 +1,31 @@
...
+	@include scss.block-properties("share-bar-envelope");
+	@include scss.block-components("share-bar-envelope");
```
matthewroach: Can we be consistent in the naming please

```suggestion
	@include scss.block-properties("share-bar-email");
	@include scss.block-components("share-bar-email");
```

blocks/share-bar-block/features/share-bar/default.jsx
```diff
@@ -135,7 +120,17 @@ export const ShareBar = ({
 		return null;
 	}
 
-	return <div className="ts-share-bar">{shareButtons}</div>;
+	return (
+		<Stack
+			className={BLOCK_CLASS_NAME}
+			alignment="center"
+			direction="vertical"
+			gap="8px"
```
matthewroach: Gap should be set at the theme file level

```suggestion
```

.storybook/themes/news.scss
```diff
@@ -685,6 +685,41 @@
...
+				share-bar-linkedin: (
+					fill: #3076b0,
+				),
+				share-bar-pinterest: (
+					fill: #ad2526,
+				),
+				share-bar-twitter: (
+					fill: #6baae8,
+				),
+				share-bar-envelope: (
```
matthewroach: As per suggestion in the CSS

```suggestion
				share-bar-email: (
```

.storybook/themes/news.scss
```diff
@@ -685,6 +685,41 @@
...
+					components: (
+						button-medium: (
+							background: transparent,
+							border-width: 0,
+							padding: 0 10px,
+							text-align: center,
+							width: 44px,
+						),
+					),
+				),
```
matthewroach: We should not be setting explicit `width` values if we don't need to, this allows the contents to grow responsively 

```suggestion
				share-bar: (
					background: var(--background-color),
					box-shadow: var(--global-box-shadow-1),
					border-radius: 0 var(--border-radius) var(--border-radius) 0,
					display: none,
					gap: var(--global-spacing-2),
					left: 0,
					padding: var(--global-spacing-4) 0,
					position: fixed,
					top: 200px,
					components: (
						button-medium: (
							background: transparent,
							border-width: 0,
							padding: 0 var(--global-spacing-3),
						),
					),
				),
```

matthewroach (changes requested):

.storybook/themes/news.scss
ecarlisle: @JackHowa : Got it. No longer using a height. Thanks!

blocks/share-bar-block/README.md
ecarlisle: Gotcha! Updated. Thanks!

.storybook/themes/news.scss
```diff
@@ -685,6 +685,39 @@
 				right-rail-advanced-main-right-rail-bottom: (
 					grid-area: main-right-rail-bottom,
 				),
+				share-bar: (
+					background: var(--background-color),
+					box-shadow: var(--global-box-shadow-1),
+					border-radius: 0 var(--border-radius) var(--border-radius) 0,
+					display: none,
```
JackHowa, line 693: this might be a dumb idea: but if there's display none for default, would the `desktop` breakpoint have all of the styles anyway? Would this reduce css duplication?

matthewroach: You are not being dumb - as it's mobile first the styles are applied here and we just don't display the block - the desktop toggles the display value

matthewroach (approved):

JackHowa (commented):
I have one performance suggestion of moving the default styles to the desktop styles. Basically, the default option (mobile) is display none. So all of its other styles are moot. These styles like background should be moved to desktop breakpoint only to reduce unused styles.

Otherwise this is good to go. Good work addressing feedback and working with me

matthewroach:
@JackHowa Moving the styles to desktop only actually increases the styles needed and has a performance impact, due to the mobile first approach and how the library works if you set styles on desktop only you have to define them on mobile to ensure they are picked up

ecarlisle:
> I have one performance suggestion of moving the default styles to the desktop styles. Basically, the default option (mobile) is display none. So all of its other styles are moot. These styles like background should be moved to desktop breakpoint only to reduce unused styles.
> 
> Otherwise this is good to go. Good work addressing feedback and working with me

@JackHowa : I appreciate the feedback. Initially this was how I had the styles set up, with the `default()` block property values set to initial. I changed as per this feedback (cc: @matthewroach ):

https://github.com/WPMedia/arc-themes-blocks/pull/1365#discussion_r886914845

 There are pros and cons of each approach:

For setting all desktop styles in `desktop()`:
**Pro:** CSS is arguably more self-documenting, showing at which breakpoint changes are made.
**Con:** CSS is bigger with the need to set property values to `initial` at `default()`.

For setting all desktop styles in `default()`, we have the opposite:
**Pro:** Smaller CSS
**Con:** Arguably less intuitive.

It's definitely worth more discussion if we're able to overcome the `initial` value requirement for blocks.

PR #1376, opened 2022-06-10
Add additional stories for header nav chain

matthewroach:
Add additional stories for Header Nav Chain to allow for using chromatic more with v2 migration


JackHowa (approved):
Good approach -- should help brent

PR #1388, opened 2022-06-20
WIP feedback Tmedia 747 feedback small manual promo

JackHowa:
- @ecarlisle testing out how we can mimic previous behavior modelling off of the Numbered List. Let me know if this makes sense 

before 
Screen Shot 2022-06-20 at 11 14 12

after 
Screen Shot 2022-06-20 at 11 13 32

### DIFF
--- a/.storybook/themes/news.scss
+++ b/.storybook/themes/news.scss
@@ -1059,7 +1059,7 @@
 				small-manual-promo: (
 					components: (
 						media-item: (
-							max-width: 33%,
+							flex-basis: 33%,
 						),
 						image: (
 							display: inline,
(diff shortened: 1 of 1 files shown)


.storybook/themes/news.scss
```diff
@@ -1059,14 +1059,14 @@
 				small-manual-promo: (
 					components: (
 						media-item: (
-							flex: 1 0 min-content,
+							max-width: 33%,
```
nschubach: This won't work on the "above"/"below" states... perhaps:
```suggestion
							flex-basis: 33%,
```
(untested)

JackHowa: ah nice -- that makes sense. I was only looking at the horizontal ones 🤦

PR #1398, opened 2022-06-29
Additional test coverage

nschubach:
Added test coverage


blocks/card-list-block/features/card-list/default.jsx
```diff
@@ -42,7 +42,7 @@ const CardListItems = (props) => {
 			title = "",
 			offsetOverride = 0,
 			displayAmount,
-		} = {},
+		},
```
nschubach, line 45: There should never be no customFields for blocks, right?

matthewroach (approved):

JackHowa (approved):
big thanks Nick

PR #1414, opened 2022-07-12
Update center logo story to have logo in center

matthewroach:
center logo story is misconfigured for center logo showing in the center - this updates the story to be correct as per pagebuilder rendering


JackHowa (approved):
Looks good, makes sense

PR #1420, opened 2022-07-14
Sync news theme as per feature pack repo

matthewroach:


JackHowa (approved):
looks good with the carousel buttons

PR #1421, opened 2022-07-14
Updates to tokens to address component regressions

matthewroach:
Same as updates applied on components - https://github.com/WPMedia/arc-themes-components/pull/116


JackHowa (approved):
good work

PR #1411, opened 2022-07-11
Tmedia 741 large promo block

JackHowa:
Help wanted to complete the promo as soon as possible. My progress so far: 

- [x] remove Index.js
- [x] update package.json
- [x] update documentation
- [x] Add _index.scs
- [x] Wire up styles 
- [x] Take in image data
- [x] Take in description data
- [x] Take in headline data
- [x] Take in overline data
- [x] Style using grid in large manual promo
- [x] Remove and refactor based on other promos the videoplayercustomfields
- [x] implement show promo label with like gallery icon. if content has a label for gallery, it looks it up in the content object to show 
- [x] Add Video and Play in Place logic. Favors a video with play video in place over an image
- [x] Add a storybook story with video and play in place
- [x] Ensure searchable image logic still works in pb
- [x] [Update translation phrases to be in either block intl.json or global phrases](https://github.com/WPMedia/arc-themes-blocks/pull/1431)
- [x] Verify removing all imports not needed and replace 
- [ ] Upon review approval, make sure https://github.com/WPMedia/arc-themes-manifests/pull/96 is marked as ready for review
- [x] Create a feature pack update with the same styles as in this themes styles
**If you have not filled out the checklist below, the pr is not ready for review.**

# Onboard 

- `npm run storybook`

## Description

Migrate the Promo blocks to the new architecture. This includes Extra Large, Large, Medium and Small Promo blocks. This ticket covers the Large Promo block.

## Jira Ticket

- [TMEDIA-741](https://arcpublishing.atlassian.net/browse/TMEDIA-741)

## Acceptance Criteria

- Large Promo block is migrated to the new architecture.

- Dividing lines between items in the block are included, but any dividing lines framing the block will be done at the layout level

## Test Steps

- Add test steps a reviewer must complete to test this PR

1. Checkout this branch `git checkout TMEDIA-741-large-promo-block
2. Run fusion repo with linked blocks npx fusion start -f -l @wpmedia/large-promo-block
3. Open https://localhost/pagebuilder/pages and open Reference - All Blocks
4. Configure Large Promo Block to use _id : 7TLCY4UZN5F4BA5I4FP62E6NYE

## Author Checklist

_The author of the PR should fill out the following sections to ensure this PR is ready for review._

- [x] Confirmed all the test steps a reviewer will follow above are working.
- [x] Confirmed there are no linter errors. Please run `npm run lint` to check for errors. Often, `npm run lint:fix` will fix those errors and warnings.
[...]

PR #1431, opened 2022-07-18
Tmedia 741 byline global phrase

JackHowa:
- Move translations in lokalise from byline to global phrases 
- Run generate intl to generate the block changes
- goes with @malavikakoppula 's large promo #1411


blocks/overline-block/intl.json
```diff
@@ -1,13 +1,13 @@
 {
 	"overline-sponsored-content": {
-		"de": "Werblicher Inhalt",
+		"de": "Sponsored Content",
```
JackHowa: interesting this looks like a regression

blocks/byline-block/intl.json
```diff
@@ -1,22 +0,0 @@
-{
```
matthewroach, line 1: Removing this file will require updates to the code in this block to ensure it's pull the translation by the new key names now

blocks/shared-styles/intl.json
```diff
@@ -21,28 +21,6 @@
 		"pt": "Vídeo",
 		"sv": "Video"
 	},
-	"byline.by-text": {
```
matthewroach: This will mean current blocks (not migrated) will not have any translations for these phrases at the moment

JackHowa: oh I thought that's what we wanted. will update

matthewroach: Yeah - that's fine by me, but may cause some confusion when a developer comes to do a migration of a block still using code from Shared Styles - technically there are still consumers of this code until the last block is converted

blocks/byline-block/intl.json
JackHowa: updated

blocks/shared-styles/intl.json
JackHowa: added back

JackHowa:
interesting -- identity seems to have changed in lokalise. looks like something with connected as changed. it should just be "connect"

JackHowa:
hopefully this will resolve @nschubach and @malavikakoppula's work

blocks/byline-block/intl.json
matthewroach: I pushed an updated - as the byline block intl.json file is manually curated and the values are not part of any locale main file

JackHowa: nice

blocks/overline-block/intl.json
matthewroach: We can take the existing values and add to lokalise then regenerate the locale files to avoid this regression

matthewroach (dismissed):
We need to update locale to avoid the sponsored content regression

nschubach:
> We need to update locale to avoid the sponsored content regression

@matthewroach : Are you referring to this which I added to `extra-large-promo-block`?
https://github.com/WPMedia/arc-themes-blocks/blob/f67cf98c4dc14f5d0e46ffa2a81fc00d36132f5d/blocks/global-phrases-block/intl.json#L134-L144

matthewroach:
@nschubach - Yes, if we have those translation available we should be setting them in their respective locations correctly

blocks/overline-block/intl.json
matthewroach: We need to fix these before merging

JackHowa: oh ok

JackHowa: it seems like lokalise may have an issue ... trying to see what's happened there

JackHowa: cool can manually update lokalise with this

JackHowa: updated to overline

JackHowa:
@matthewroach feedback around sponsored content has been addressed. not sure how that happened on lokalise's side. updated to global phrases

blocks/subscriptions-block/intl.json
```diff
@@ -77,15 +77,15 @@
...
-		"sv": "Connected as %{email}"
+		"de": "Connected",
+		"en": "Connected",
+		"es": "Connected",
+		"fr": "Connected",
+		"ja": "Connected",
+		"ko": "Connected",
+		"no": "Connected",
+		"pt": "Connected",
+		"sv": "Connected"
```
nschubach, line 88: You said this broke something?

It's passed in Subs, ContactInfo: 
image

nschubach: not sure it's necessary as is so I don't care either way.

locale/de.json
```diff
@@ -1,448 +1,447 @@
...
+		"category-carousel.left-arrow-label": "Previous",
+		"category-carousel.slide-indicator": "Slide %{current} of %{maximum}"
+	},
+	"product-assortment-carousel-block": {
+		"product-assortment-carousel.aria-label": "Products",
+		"product-assortment-carousel.right-arrow-label": "Next",
+		"product-assortment-carousel.left-arrow-label": "Previous",
+		"product-assortment-carousel.slide-indicator": "Slide %{current} of %{maximum}"
+	}
+}
```
nschubach:
```suggestion
}

```

nschubach (approved):

PR #1438, opened 2022-07-21
Docs themes setup feedback

JackHowa:
Feedback from internal use: 

- How to setup the blocks storybook in the readme
- Outdated documentation around lerna and releasing
- update the intro on the blocks storybook intro
- also add a date of change to the intro
- updated based on feedback from @annasherman and @jgrosspietsch -- thank you 🆒


matthewroach (approved):
Looks good to me - I removed the two readme templates as they are now part of the generators

PR #1440, opened 2022-07-21
TBRANDS-96 - Update CSS and Heading

matthewroach:
* Adding in the missing line height that was applied to the feature pack
* Removed the `HeadingSection` as Production Name should be a `h1` on the page


JackHowa (approved):
good to go

PR #1468, opened 2022-08-04
Tmedia 741 include label

JackHowa:
- Add storybook way of accounting for different types of content 
- Ensure that play in place still works as expected
- Add todos for lead art content 
- Test breaking for icon label


malavikakoppula (approved):
TYVM

PR #1499, opened 2022-08-31
translations: migrate full author and other changes

JackHowa:
- Adding back translations


JackHowa:
- translations: 

goal: 
- in lokalise, keep v1 translations for maintainability
- in v2 code, remove v1 translations for performance 
- in v2 code, use the global.social-{service}-connect for handling phrases in short and full author bio 
- @nschubach will merge this into his pr to unblock his work 
- Same with malavika in the short author bio to unblock
- we don't need social-{service}-content in global phrases so that was deleted in lokalise as welll

malavikakoppula (commented):
news.scss has large-promo styles, which had been already merged into `arc-themes-release-version-2.0.1`

PR #1501, opened 2022-08-31
Tmedia 734 gallery block single line top wip

JackHowa:
- Single line at the top

todo: 
- [ ] Ensure no side effects of defining a component from within another block
- [ ] Figure out how design wants to handle longer text here

Screen Shot 2022-08-31 at 17 32 43


JackHowa:
not needed based on dqa

PR #1519, opened 2022-10-03
ci: Update the package-lock to use the correct dependencies in the lockfile

JackHowa:
- Update lockfile so that the many false security warnings are reduced. 
- The current lockfiles were also not representative on the package.json
- I ran `npx lerna exec --no-bail --parallel -- npm i` to generate the lockfile or update it

false dependabot updates 

Screen Shot 2022-10-03 at 14 49 04
Screen Shot 2022-10-03 at 14 49 00


matthewroach (approved):

JackHowa:
working as expected 

Screen Shot 2022-10-04 at 10 18 03

PR #1531, opened 2022-10-14
tmedia 962: Fix typo to trigger re-render after botched publish of resizer block

JackHowa:
- Publish didn't work before so changed block needs to be published https://github.com/WPMedia/arc-themes-blocks/actions/runs/3237014216/jobs/5303551315
- Fixed small typo


JackHowa (commented):
Small change to trigger. wanted to document the update

PR #1534, opened 2022-10-17
Docs on generators actions linting

JackHowa:
# Goals

- [x] Document and audit all existing automation in a readme in plain dev language
- [x] Document hygen generation and available commands
- [x] Document how github actions work with lerna
- [x] Discuss storybook


.storybook/manager-head.html
```diff
@@ -1,44 +1 @@
 <!-- reference for customizing https://github.com/storybookjs/storybook/issues/11222#issuecomment-646411961 -->
-<link rel="stylesheet" href="https://nav-prod.home.static.arcpublishing.com/css/arcnav.css" />
```
matthewroach: There are no current plans to not render this in ALC going forward - while that could change, we should keep this until we decided either way

JackHowa: ok will revert

JackHowa: updated

matthewroach (approved):
Thanks for the clean up @JackHowa

PR #1539, opened 2022-10-20
Add data-chromatic='ignore' to image in header nav

matthewroach:
## Description

We've been seeing a lot of false positives with our chromatic runs for the header nav image not always loading, we can ignore element in chromatic with the `data-chromatic="ignore"` HTML attribute. This adds that attribute to the header nav image to see if we can reduce the false positives during our test runs

**Update**: Added the logo's to our static directory `resources` to avoid relying on a third party service for rendering - this will also help reduce false positives with images loading


JackHowa (dismissed):
good idea

JackHowa (approved):
makes sense, good work

PR #1540, opened 2022-10-25
Tmedia 934

nschubach:
## Description

Convert `article-body-block` to image resizer.
Also fix condition in `content-api` that I missed.

## Jira Ticket

- [TMEDIA-934](https://arcpublishing.atlassian.net/browse/TMEDIA-934)

## Acceptance Criteria

Article Body block is using image resizer v2

## Test Steps

1. Checkout this branch `git checkout tmedia-934`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/article-body-block`
3. Open the Article Body Chain test page and use content-api countent source with id:  `PKFVHJKBBVHZRIHJEMQGGREWIE`
4. Check that the appropriately sized image is loaded and that the markup is as below (in the After section)

### Before

image

### After

image

## Author Checklist

## Reviewer Checklist

### DIFF
--- a/blocks/article-body-block/chains/article-body/default.jsx
+++ b/blocks/article-body-block/chains/article-body/default.jsx
@@ -49,13 +49,13 @@ function parseArticleItem(item, index, arcSite, phrases, id, customFields) {
 	switch (type) {
 		case "text": {
 			return content && content.length > 0 ? (
-				<Paragraph key={`${type}_${key}`} dangerouslySetInnerHTML={{ __html: content }} />
+				<Paragraph key={`${type}_${index}_${key}`} dangerouslySetInnerHTML={{ __html: content }} />
 			) : null;
 		}
 		case "copyright": {
 			return content && content.length > 0 ? (
 				<Paragraph
-					key={`${type}_${key}`}
+					key={`${type}_${index}_${key}`}
 					className={`${BLOCK_CLASS_NAME}__copyright`}
 					dangerouslySetInnerHTML={{ __html: content }}
 				/>
@@ -63,7 +63,7 @@ function parseArticleItem(item, index, arcSite, phrases, id, customFields) {
 		}
 
 		case "divider": {
-			return <hr className={`${BLOCK_CLASS_NAME}__divider`} key={`${type}_${key}`} />;
+			return <hr className={`${BLOCK_CLASS_NAME}__divider`} key={`${type}_${index}_${key}`} />;
 		}
 
 		case "image": {
@@ -89,7 +89,7 @@ function parseArticleItem(item, index, arcSite, phrases, id, customFields) {
 				const formattedCredits = formatCredits(vanityCredits || credits);
 				return (
 					<MediaItem
-						key={`${type}_${key}`}
+						key={`${type}_${index}_${key}`}
 						className={figureImageClassName}
 						caption={!hideImageCaption ? caption : null}
 						credit={!hideImageCredits ? formattedCredits : null}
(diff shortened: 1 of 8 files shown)


blocks/article-body-block/chains/article-body/default.jsx
```diff
@@ -218,7 +219,7 @@ function parseArticleItem(item, index, arcSite, phrases, id, customFields) {
 		case "gallery": {
 			const total = item.content_elements.length;
```
matthewroach, line 223: We also need to update the `<Image />` components in the galleries to use resizer

blocks/article-body-block/chains/article-body/default.test.jsx
```diff
@@ -81,7 +90,7 @@ describe("article-body chain", () => {
 					<span>3</span>
 				</ArticleBodyChain>
 			);
-
+			//
```
matthewroach:
```suggestion

```

blocks/article-body-block/chains/article-body/default.jsx
```diff
@@ -44,84 +49,66 @@ function parseArticleItem(item, index, arcSite, phrases, id, customFields) {
...
 						title={!hideImageTitle ? subtitle : null}
 					>
-						{children}
+						<Conditional component={Link} condition={link} href={link}>
+							<Image
+								src={imageANSToImageSrc(item)}
+								alt={altText}
+								resizerURL={RESIZER_URL}
+								resizedOptions={{ auth: auth[RESIZER_APP_VERSION] }}
+								width={item.width}
```
matthewroach: Do we want to use the raw `width` and `height` here? For the testing article I am seeing 3024x4032

matthewroach (changes requested):
Couple small comments, and we need to ensure we include Gallery images in the update too

blocks/article-body-block/chains/article-body/default.jsx
nschubach: I wanted to keep the aspect ratio and I suppose I could do some math to reduce it to something more in line, but with the responsive resizing it could be any value in a range.  I thought about probably figuring out the max width (which will be different depending on if it's floating...) and then calculate the height.  If I did this though, we'd want to be able to do this anywhere image is used and it's getting to the point where we want yet another util...

nschubach:
I also noticed I need to figure out mocking on the storybook image url to return the image url we have just so it's not broken images everywhere (since the resizer uses the resizer url.

blocks/article-body-block/chains/article-body/default.jsx
nschubach: Latest update accounts for the gallery as well.

blocks/article-body-block/chains/article-body/default.jsx
nschubach: I broke out a method in this to hopefully manage this a little better.

nschubach:
Fix one thing, break another... going to resolve after lunch.

blocks/article-body-block/chains/article-body/default.jsx
```diff
@@ -1,22 +1,27 @@
...
 	Image,
+	isServerSide,
 	Link,
 	MediaItem,
 	Paragraph,
 	usePhrases,
 	Video,
 } from "@wpmedia/arc-themes-components";
 
+import getResizeParamsFromANSImage from "./shared/get-resize-params-from-ans-image";
```
matthewroach, line 23: Seems like a good candidate to move the function to Arc Themes Components? We can create a ticket to do this at a later stage if we start to see a pattern of needing to do the same logic across other blocks

matthewroach (approved):
Looks good to me

PR #1619, opened 2023-02-08
THEMES-965: Add themes=v2 parameter to each used content source

nschubach:
## Description

With the updates for Enhanced Styling and Resizer V2 - Content Sources have been converted all to use the fetch method and also have a different response contract. In order for deployments to themes v2 to no cause issues for clients by using existing cached responses and arbitrary param needs added to all content sources to ensure a cached content source is not returned

## Jira Ticket

- [THEMES-965](https://arcpublishing.atlassian.net/browse/THEMES-965)

## Acceptance Criteria

For all content sources add a new param named themes with a default value of v2

## Test Steps

1. Checkout this branch `git checkout themes-965`
2. Run fusion repo with linked blocks `npx fusion start -f -l`
3. Testing this is a bit difficult because the API is called on the back end and only visible in the fusion console and look for the `&themes=v2` appended to the url request:
```
https://api.sandbox.themesinternal.arcpublishing.com/content/v4/search/published?body=%7B%22query%22%3A%7B%22bool%22%3A%7B%22must%22%3A%5B%7B%22term%22%3A%7B%22revision.published%22%3A%22true%22%7D%7D%2C%7B%22nested%22%3A%7B%22path%22%3A%22taxonomy.sections%22%2C%22query%22%3A%7B%22bool%22%3A%7B%22must%22%3A%5B%7B%22terms%22%3A%7B%22taxonomy.sections._id%22%3A%5B%22%2Fnews%22%5D%7D%7D%2C%7B%22term%22%3A%7B%22taxonomy.sections._website%22%3A%22the-gazette%22%7D%7D%5D%7D%7D%7D%7D%5D%2C%22must_not%22%3A%5B%7B%22nested%22%3A%7B%22path%22%3A%22taxonomy.sections%22%2C%22query%22%3A%7B%22bool%22%3A%7B%22must%22%3A%5B%7B%22terms%22%3A%7B%22taxonomy.sections._id%22%3A%5B%22%22%5D%7D%7D%2C%7B%22term%22%3A%7B%22taxonomy.sections._website%22%3A%22the-gazette%22%7D%7D%5D%7D%7D%7D%7D%5D%7D%7D%7D&from=2&size=6&sort=display_date%3Adesc&website=the-gazette&themes=v2
```

## Author Checklist

## Reviewer Checklist

[THEMES-965]: https://arcpublishing.atlassian.net/browse/THEMES-965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/blocks/alert-bar-content-source-block/sources/alert-bar-collections.js
+++ b/blocks/alert-bar-content-source-block/sources/alert-bar-collections.js
@@ -1,16 +1,23 @@
 import axios from "axios";
 import { ARC_ACCESS_TOKEN, CONTENT_BASE } from "fusion:environment";
 
-const params = {};
+const params = [
+	{
+		default: "2",
+		displayName: "Themes Version",
+		name: "themes",
+		type: "text",
+	},
+];
 
-const fetch = ({ "arc-site": website }) => {
+const fetch = ({ themes, "arc-site": website }) => {
 	const urlSearch = new URLSearchParams({
 		content_alias: "alert-bar",
 		from: 0,
 		published: true,
 		size: 1,
 		...(website ? { website } : {}),
-		themes: "v2",
+		themes,
 	});
 
 	return (
--- a/blocks/alert-bar-content-source-block/sources/alert-bar-collections.js
+++ b/blocks/alert-bar-content-source-block/sources/alert-bar-collections.js
@@ -10,6 +10,7 @@ const fetch = ({ "arc-site": website }) => {
 		published: true,
 		size: 1,
 		...(website ? { website } : {}),
+		themes: "v2",
 	});
 
 	return (
--- a/blocks/alert-bar-content-source-block/sources/alert-bar-collections.test.js
+++ b/blocks/alert-bar-content-source-block/sources/alert-bar-collections.test.js
@@ -35,7 +35,14 @@ jest.mock("axios", () => ({
 
 describe("the author api content source block", () => {
 	it("should use the proper param types", () => {
-		expect(contentSource.params).toEqual({});
+		expect(contentSource.params).toEqual([
+			{
+				default: "2",
+				displayName: "Themes Version",
+				name: "themes",
+				type: "text",
+			},
+		]);
 	});
 
 	it("should build the correct url", async () => {
--- a/blocks/author-content-source-block/sources/author-api.js
+++ b/blocks/author-content-source-block/sources/author-api.js
@@ -11,6 +11,7 @@ const params = {
 const fetch = ({ slug }, { cachedCall }) => {
 	const urlSearch = new URLSearchParams({
 		slug,
+		themes: "v2",
 	});
 
 	return axios({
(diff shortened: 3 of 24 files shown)


blocks/alert-bar-content-source-block/sources/alert-bar-collections.js
```diff
@@ -10,6 +10,7 @@ const fetch = ({ "arc-site": website }) => {
 		published: true,
 		size: 1,
 		...(website ? { website } : {}),
+		themes: "v2",
```
vgalatro: This is a question for all of these changes, but why are we adding the param here to the content API call? Don't we need to update the cache key for the content source itself? If so, this should be in the `params` object above. I think there's a way to set a default value for it so we don't have to update every single call to these content sources as well. I'll try to dig up the relevant docs.

vgalatro (commented):
I just have a couple of questions about the changes that were made.

blocks/alert-bar-content-source-block/sources/alert-bar-collections.js
vgalatro: Looks like the default value is only for content sources defined in JSON: https://docs.arcxp.com/alc?sys_kb_id=fdba431f476e1950a87626c2846d430c&id=kb_article_view. I guess the expectation is that you can set the defaults in the code if you're writing it out in JS. I'm now wondering what the expected change was when Matthew thought up this task. It seems like a bigger lift than necessary for something I'm 90% sure will be handled by the `d` parameter updating on the deployment being published.

nschubach: Adding them to the params is definitely possible.  It wasn't very clear to me what the point of the story was besides cache busting which could happen in two areas... one for Pagebuilder cache busting and the other for API call cache busting.  I could really add it to both for as much as this feels like a temporary solution to cover an edge case.

Thoughts @rmbrntt ?

rmbrntt: Decision point for me would be our confidence in the d= key breaking the cache, if high confidence then it sounds covered and we can keep this PR on hold but not close it (as a back-up option). If not high confidence then this PR is good to go. And worst case we just remove it after our sunset of 1.x date and all customers have moved. 

For now I'm thinking we can put this on hold, consider the ticket done in this sprint. And then keep the PR open but do not merge.

We have a dry run ticket where we should be able to test this out...

nschubach: The d param is applied to the client call, but this change applies to the serverside call that calls for the server to run this content source.  But there's a question as to this being a change to break the pagebuilder cache of the content source itself (which AFAIK is a very short term cache...)

rmbrntt (approved):

PR #1627, opened 2023-02-22
Removing SASS division values and replacing with css calc()

nschubach:
Remove the sass division values and replace with css calc()


malavikakoppula (approved):
LGTM

PR #1729, opened 2023-09-13
remove final references to styled components, remove as dependency fr…

annasherman:
Styled components dependency was causing a circular dependency issue for several clients (mismatched versions of styled components). Also, we are not actually using styled components anymore, so it should not be installable and should not cause this error.


vgalatro (approved):
LGTM!

PR #1743, opened 2023-10-11
Update arc-themes-base-to-new-tag.yml

vgalatro:
## Description

PR to test a fix for the github workflow "Tag New Arc Theme Release". This is being made on the 1.29 branch for now, to facilitate the creation of the 1.30 release.

### DIFF
--- a/.github/workflows/arc-themes-base-to-new-tag.yml
+++ b/.github/workflows/arc-themes-base-to-new-tag.yml
@@ -49,12 +49,12 @@ jobs:
 
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup Node and NPM
         uses: actions/setup-node@v3
         with:
-          node-version-file: .nvmrc
+          node-version: "16"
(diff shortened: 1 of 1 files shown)


.github/workflows/arc-themes-base-to-new-tag.yml
```diff
@@ -49,12 +49,12 @@ jobs:
...
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup Node and NPM
         uses: actions/setup-node@v3
         with:
-          node-version-file: .nvmrc
+          node-version: "16"
```
nschubach, line 57: I'll accept this, but when we move to fusion 4, I think it's fair to revert these to use the nvmrc for the version so we don't have different versions all over the place...

nschubach (approved):

.github/workflows/arc-themes-base-to-new-tag.yml
vgalatro: Yes, that was my idea too. Having just this set to 16 shouldn't break anything else in the meantime.

PR #2106, opened 2024-04-26
THEMES-1900: Fix logic for detecting 404 for sectionId/_id

nschubach:
#### Jira Ticket: [THEMES-1900](https://arcpublishing.atlassian.net/browse/THEMES-1900)

Fixing section api 404 detection

## Test Steps

1. Checkout this branch `git checkout themes-1900`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/site-hierarchy-content-block`
3. http://localhost/pagebuilder/tools/debugger?_website=the-baltimore-banner&type=content-debugger&source=site-service-hierarchy&config={%22sectionId%22:%22/asdf%22,%22hierarchy%22:%22%22}
4. also search for `/topic/education` section which should return data

[THEMES-1900]: https://arcpublishing.atlassian.net/browse/THEMES-1900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/blocks/site-hierarchy-content-block/sources/site-service-hierarchy.js
+++ b/blocks/site-hierarchy-content-block/sources/site-service-hierarchy.js
@@ -11,11 +11,11 @@ export default {
 		sectionId: "text",
 	},
 	transform: (data, query) => {
-		if (query?.sectionId === data._id) {
-			return data;
+		if (query.sectionId && query.sectionId !== data._id) {
+			const error = new Error("Not found");
+			error.statusCode = 404;
+			throw error;
 		}
-		const error = new Error("Not found");
-		error.statusCode = 404;
-		throw error;
+		return data;
 	},
 };
--- a/blocks/site-hierarchy-content-block/sources/site-service-hierarchy.js
+++ b/blocks/site-hierarchy-content-block/sources/site-service-hierarchy.js
@@ -1,27 +1,21 @@
 export default {
-	resolve(resolveParams) {
-		const { hierarchy, sectionId, "arc-site": arcSite } = resolveParams;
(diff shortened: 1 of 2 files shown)


blocks/site-hierarchy-content-block/sources/site-service-hierarchy.js
```diff
@@ -11,17 +11,11 @@ export default {
 		sectionId: "text",
 	},
 	transform: (data, query) => {
-		let idMatch = false;
-		if (query.sectionId) {
-			idMatch = data._id !== query.sectionId;
+		if (query?.sectionId === data._id) {
```
vgalatro: This also needs to check if the `sectionId` was queried in the first place. This change breaks the section nav and footer since using this source for just a nav hierarchy always 404's.

Using the same logic as v2 will probably work:
```
if (query.sectionId && query.sectionId !== data._id) {
				const error = new Error("Not found");
				error.statusCode = 404;
				throw error;
}

return data;
```

vgalatro (changes requested):
This change breaks blocks like the header nav chain and footer that only query this content source for a nav hierarchy without a section ID set.

blocks/site-hierarchy-content-block/sources/site-service-hierarchy.js
nschubach: `query?.sectionId` is that check?

nschubach: oh, nm, I see what you are saying...

vgalatro (approved):
LGTM!

PR #2132, opened 2024-05-16
[Themes 1930] Added height to header-nav-chain-flyout-overlay-open

malavikakoppula:
#### Jira Ticket: [THEMES-1930](https://arcpublishing.atlassian.net/browse/THEMES-1930)
We noticed that there’re some items of the side menu navigation list which never reach to be visible on the screen after scrolling the side menu down.
I attach some screenshots to clarify my description

Once the side menu is scrolled down “to the end of it”, the entire screen starts to scroll like there was no more items to show, but this is no true. 
It appears that the height of the side menu is not being properly calculated.

## Test Steps

1. Checkout this branch `git checkout THEMES-1930`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/header-nav-chain-block`
3. Flyout overlay should be scrollable when opn

[THEMES-1930]: https://arcpublishing.atlassian.net/browse/THEMES-1930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.storybook/themes/news.scss
+++ b/.storybook/themes/news.scss
@@ -1636,7 +1636,8 @@
 					"overflow-y": scroll,
 					"overflow-block": scroll,
 					"transform": translate(0, 0),
-					"inline-size": 100%
+					"inline-size": 100%,
+					"height": calc(100vh - var(--header-nav-chain-height))
 				),
 				"header-nav-chain-flyout-overlay-scrollbar": (
 					"display": none
--- a/.storybook/themes/news.scss
+++ b/.storybook/themes/news.scss
@@ -1636,8 +1636,7 @@
 					"overflow-y": scroll,
 					"overflow-block": scroll,
 					"transform": translate(0, 0),
-					"inline-size": 100%,
-					"height": calc(100vh - var(--header-nav-chain-height))
+					"inline-size": calc(100vh - var(--header-nav-chain-height))
 				),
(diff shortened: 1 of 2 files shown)


.storybook/themes/news.scss
```diff
@@ -1636,7 +1636,8 @@
 					"overflow-y": scroll,
 					"overflow-block": scroll,
 					"transform": translate(0, 0),
-					"inline-size": 100%
+					"inline-size": 100%,
+					"height": calc(100vh - var(--header-nav-chain-height))
```
nschubach: We should update the `inline-size` we don't set height with logical properties now.

nschubach (changes requested):
We should be using logical properties.

.storybook/themes/news.scss
malavikakoppula: ah! I forgot

malavikakoppula: @nschubach Fixed it in [d6be774](https://github.com/WPMedia/arc-themes-blocks/pull/2132/commits/d6be774721be7681b9e216aa0856f20fceea8ac6)

nschubach (approved):

PR #2138, opened 2024-05-21
Bug/asub 8456 reset password login redirect issue

accbjt:

## Description
Fix for Irish News redirect from reset password issue. When selecting redirect to previous page, after login the user would be redirected to the reset password page they were on so it looks like they never reset their password and logged in. This will redirect them to the original redirect path.

#### Jira Ticket: [THEMES-](https://arcpublishing.atlassian.net/browse/THEMES-)
https://arcpublishing.atlassian.net/browse/ASUB-8456
_Information about what you changed for this PR. Include any dependencies, companion PRs, relevant screenshots, or config changes._

## Test Steps
- Add Reset Password URL to the login page. Make sure it is coming from the correct reset password page.
Screenshot 2024-05-20 at 10 03 45 PM
- Make sure you also have a Redirect URL as well.
- Go to your reset password page and reset your password. It will prompt you to login after.
- Click login and then login and it should redirect you to the original Redirect URL and not the previous page you were on.

1. Checkout this branch `git checkout {branch name}`
2. Run fusion repo with linked blocks `npx fusion start -f -l {blocks to link}`
3. {Next ....}


LauraPinilla:
Hi @accbjt 
The fix looks good, However, I think we need to append the redirect param when we click on continue once the nonce was redeemed successfully, otherwise, when we fall on the login page we will see the same behavior since the resetPasswordURL is not part of the URL search params.

https://github.com/WPMedia/arc-themes-blocks/assets/54566275/03a859e7-a6d7-47da-85ff-8864dc98ff85

By checking this functionality I saw a few things that maybe we need to discuss with @mattkim93 and Jessica in order to improve the block functionality. It's not related directly to the ticket description, but we should avoid our clients came back reporting new issues related to this feature.

1. When we fall on the page where we have the `Identity-reset-password` block, if the nonce doesn't exist, we are showing a blank page. Maybe in those cases, the user should be redirected to the main page or a different page. 

https://github.com/WPMedia/arc-themes-blocks/assets/54566275/5964341d-afea-4612-b3ad-12d68c13c5ed

2. Also I can see we are providing a general error message, we should provide a better error message, and if the BE returns `{"httpStatus":400,"code":"300033","message":"Nonce not found or expired"}` a button that allows us go back into the reset-password page will be great. 
Screenshot 2024-05-23 at 9 39 37 p m

Let me know if you have questions.

blocks/identity-block/features/reset-password/default.jsx
```diff
@@ -50,6 +55,16 @@ export const ResetPasswordPresentation = ({ isAdmin = false, phrases, successAct
 		}
 	}, [Identity]);
 
+	const getRedirectUrl = () => {
+		const redirect = validateURL(successActionURL);
+
+		if (redirect.includes('?')) {
+			return `${redirect}&reset_password=true`
```
LauraPinilla, line 72: Thanks @accbjt I was talking about this.

LauraPinilla (dismissed):
lgtm!

blocks/identity-block/utils/validate-redirect-url.js
```diff
@@ -1,10 +1,18 @@
-const validateURL = (url) => {
+const validateURL = (url, contextPath = '') => {
 	if (!url) return null;
 	const validationRegEx = /^\/[^/].*$/;
 	const valid = validationRegEx.test(url);
+
 	if (valid) {
+		return `${window.location.origin}${contextPath}${url}`;
```
accbjt: This was added to fix the issue with AppSec requirements with out having to do too much with configurations. `contextPath` in local will append `/pf` to the url. We will have to test this out on the remote server because I am not sure what that value is on the remote server.

blocks/identity-block/components/login/index.jsx
```diff
@@ -48,16 +51,24 @@ const useLogin = ({
...
 
 		if (redirectToPreviousPage && document?.referrer) {
-			const redirectUrl = new URL(document.referrer);
+			const redirectUrlLocation = new URL(document.referrer);
 
-			setRedirectToURL(`${redirectUrl.pathname}${redirectUrl.search}`);
+			if (searchParams.has('reset_password')) {
+				setCurrentRedirectToURL(`${redirectURL}${redirectUrlLocation.search}`);
+			} else {
+				const newRedirectUrl = redirectUrlLocation.pathname.includes('/pagebuilder/')
```
accbjt, line 62: Added this to handle when we are on the edit page in pagebuilder and we go to the preview page. Since it has a referrer it will redirect back to the edit page if we do not have this logic.

blocks/identity-block/utils/validate-redirect-url.js
```diff
@@ -1,10 +1,18 @@
...
+	}
+
+	const urlLocation = new URL(url);
+
+	if (urlLocation === window.location.origin) {
```
accbjt: This is needed because it seems like we do a double validation in different areas of the code so it checks if the window origin matches the url origin.

blocks/identity-block/utils/validate-redirect-url.js
```diff
@@ -2,9 +2,22 @@ const validateURL = (url) => {
 	if (!url) return null;
 	const validationRegEx = /^\/[^/].*$/;
 	const valid = validationRegEx.test(url);
+
 	if (valid) {
+		// handling preview with PB.
+		if (window.location.pathname.includes('/pf/')) {
```
LauraPinilla: Hi @Bill, Everything looks good, but do you mind to check if there is another way (existing method) to know if we are in preview mode instead of check the URL contains `/pf/`

LauraPinilla (commented):
Hi @accbjt, by testing this manually, I found something that looks like a bug. At the end when the user is getting redirected to the login page we have double /pf/ on preview mode

https://github.com/WPMedia/arc-themes-blocks/assets/54566275/f4de35e5-e7dd-49b9-8e68-2c6be875e1cd

LauraPinilla (approved):
thanks @accbjt LGTM!

blocks/identity-block/components/login/index.jsx
vgalatro: `isAdmin` is already being passed to this hook. That should be used to prevent this redirect from happening in the first place. Something like `if (redirectToPreviousPage && document?.referrer && !isAdmin)`

vgalatro (changes requested):
There's some logic that can be changed to prevent redirects in the editor interface. I  made a comment on the relevant line.

accbjt:
@vgalatro I tested it with `!isAdmin` and it does not work on the preview page. isAdmin comes back false when I am on the login page in preview. So it ends up redirecting back to the PB page if that was the referrer.

blocks/identity-block/features/login/default.jsx
```diff
@@ -29,11 +29,21 @@ export function definedMessageByCode(code) {
...
-	const url = new URL(urlString);
+	let urlString = '';
+	let url = '';
+
+	if (window?.location?.href) {
+		urlString = window.location.href;
+		url = new URL(urlString);
+	}
 
 	const { isAdmin, arcSite } = useFusionContext();
```
vgalatro, line 48: @accbjt can you make sure that `isAdmin` is `true` here when you are on the preview page in your test environment? This is how Fusion engine identifies when it's being rendered in the the PB Editor, and it should always be true in that case. If it's not then there's a bug somewhere in Engine.

vgalatro: And to be clear, when you say "preview" do you mean viewing it the PB Editor?

accbjt: No not on PB editor directly but when you click on the eye icon to preview the page. Here is the below link to the preview page.

http://localhost/pf/login/?_website=the-gazette&outputType=default

vgalatro: Okay, I mis-understood the point of that logic then. It'll need to be there then, there's not really any way around checking the referrer URL in that specific case. I'll approve this and add notes for the CodeQL alerts that they'll be addressed in that new ticket you made.

vgalatro (approved):

PR #2159, opened 2024-06-03
Revert "THEMES-1916: Update Author images to use signed resizer image…

nschubach:
reverting change that should not have gone in 2.4


blakeganderson (approved):
LGTM

PR #2163, opened 2024-06-05
THEMES-1916: Full Author was not accounting for "original" author data

nschubach:
This is a correction for an issue found in testing.  The full-author-bio-block was not correctly identifying images since it was using the additiona_properties.original author data which is not scanned by the signing service.

## Test Steps

1. Checkout this branch `git checkout themes-1916-2`
2. Run fusion repo with linked blocks `npx fusion start -f -l`
3. Setup a page with content-api: `/2022/06/01/what-is-an-article/`and both author blocks.
image
4. Verify both author images have resizer information.

[THEMES-1916]: https://arcpublishing.atlassian.net/browse/THEMES-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blocks/full-author-bio-block/features/full-author-bio/_children/Presentation/index.jsx
```diff
@@ -59,41 +59,41 @@ const AuthorImage = ({ image, altText = "" }) => {
 	return null;
 };
 
-const Presentation = ({ author = {}, authorProfileLink }) => {
+const Presentation = ({ author }) => {
 	const phrases = usePhrases();
+
+	const originalAuthor = author?.additional_properties?.original;
```
vgalatro: The field `additional_properties.original` does not exist on author data returned from the Author API. So this breaks the block (other than the image) on the author info templates:
Screenshot 2024-06-06 at 3 14 08 PM

You can test this by setting the global content in PB Editor to use the `author-api` and a slug of `tariq-alshwaiki`.

vgalatro (changes requested):
Noticed this causes issues on the author template pages. Once that's fixed this should be good to go.

blocks/full-author-bio-block/features/full-author-bio/_children/Presentation/index.jsx
nschubach: It would have never worked that way but I will make some changes to allow it to work that way.

vgalatro (approved):
LGTM!

PR #2174, opened 2024-07-01
Remove polyfill script tag for Hotfix

vgalatro:
## Description

Removes the polyfill script tag for the default output type block.

## Test Steps

1. Checkout this branch `git checkout remove-polyfill`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/default-output-block`
3. On any page, the `polyfill.io` script tag should not be present.


malavikakoppula (approved):
LGTM

PR #2176, opened 2024-07-02
ASUB-8549 custom buttom for facebook and apple - updating user accoun…

LauraPinilla:
…t management

## Description
Updates on Social-sign-on & account management block

https://arcpublishing.atlassian.net/browse/ASUB-8521
https://arcpublishing.atlassian.net/browse/ASUB-8549

There are some AppSec Security alerts, All these will be fixed as part of https://arcpublishing.atlassian.net/browse/ASUB-8539

## Test Steps

Use the cdn endpoint https://stagingva-staging-sandbox.api.cdn.arcpublishing.com
Since this tenant has all the Third Auth Providers.

2 Blocks has been updated as part of this PR
Create a page with a Social-Sign-on block. Enable/disable the custom field **customButtons**. It will render the buttons provided by Fb or Google or the custom buttons included as part of this ticket. 
Create a page with account-management block. There you should be able to see the identities attached, or you should be able to attach new identities. 

1. Checkout this branch `git checkout ASUB-8521-SocialMediaButtons`
2. Run fusion repo with linked blocks `npx fusion start -f -l @wpmedia/identity-block`
3. Create a page with a Social-Sign-on block. Enable/disable the custom field **customButtons**. It will render the buttons provided by Fb or Google or the custom buttons included as part of this ticket. 
4 Create a page with account-management block. There you should be able to see the identities attached, or you should be able to attach/remove identities.


edwardcho1231:
Just curious but what's reasoning behind adding custom button flag on the custom fields instead of just changing the existing button? Thanks.

blocks/identity-block/components/social-sign-on/_children/GoogleSignIn.jsx
```diff
@@ -1,5 +1,54 @@
...
+
+function GoogleSignIn({ customButtons, socialSignOnIn, className }) {
+	const phrases = usePhrases();
+	const googleNotification = phrases.t("identity-block.social-google-one-tap-notification");
+
+	useEffect(() => {
+		if (customButtons) {
+			document.getElementById("custom-google-signin-btn").addEventListener("click", () =>
+				window.google.accounts.id.prompt((notification) => {
+					if (notification.isNotDisplayed() || notification.isSkippedMoment()) {
```
edwardcho1231: It seems like google doesn't want you to use `isNotDisplayed` with FedCM roll out.
https://developers.google.
Screenshot 2024-07-09 at 1 37 49 PM

[com/identity/gsi/web/guides/fedcm-migration?s=dc#display_moment](https://developers.google.com/identity/gsi/web/guides/fedcm-migration?s=dc#display_moment)

blocks/identity-block/features/account-management/_children/SocialEditableSection.jsx
```diff
@@ -1,69 +1,77 @@
...
-	const googleConnectText = phrases.t("identity-block.connect-platform", {
-		platform: "Google",
-	});
+	useEffect(() => {
+		setUpdateIdentities((prevState) => !prevState);
+	}, [checkIdentities, setUpdateIdentities]);
 
 	return (
 		<>
 			{googleClientId ? (
```
accbjt, line 44: You don't have to use a ternary and the option `null` you can just use use `&&`. That way you don't have to add the null portion of the code.

```
{googleClientId && (
 <div>Your html here</div>
)}
```

accbjt: You don't have to use a ternary and the option `null` you can just use use `&&`. That way you don't have to add the null portion of the code.

```
{googleClientId && (
 <div>Your html here</div>
)}
```

blocks/identity-block/features/account-management/_children/SocialEditableSection.jsx
```diff
@@ -1,69 +1,77 @@
...
-					<div>
-						<SocialEditableFieldContainer
-							onDisconnectFunction={unlinkFacebook}
-							text={hasFacebook ? socialText : facebookConnectText}
-							disconnectText={disconnectText}
-							showDisconnectButton={hasFacebook && hasPasswordAccount}
-						/>
-					</div>
-				</div>
+				<SocialSignOnEditableFieldContainer
```
accbjt, line 55: Same here you don't need to use the ternary.

```
{facebookAppId && (
 <div>Your html here</div>
)}
```

blocks/identity-block/features/account-management/_children/SocialEditableSection.jsx
```diff
@@ -1,69 +1,77 @@
...
+					blockClassName={blockClassName}
+					socialIcon={FacebookIcon}
+					type="Facebook"
+					isConnected={hasFacebook}
+					numIdentities={numIdentities}
+					onDisconnectFunction={unlinkFacebook}
+					updateIdentities={updateIdentities}
+				/>
+			) : null}
+			{appleTeamId && appleKeyId && appleUrlToReceiveAuthToken ? (
```
accbjt: Same here

```
{appleTeamId && appleKeyId && appleUrlToReceiveAuthToken && (
  <div>your html</div>
)}
```

blocks/identity-block/features/account-management/_children/SocialSignOnEditableContainer.jsx
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}, [type, isConnected, googleNotification]);
+
+	useEffect(() => {
+		if (type === "Facebook" && !isConnected) {
+			const attachEventListener = () =>
+				document.getElementById("custom-Facebook-signin-btn").addEventListener("click", () => {
+					window.onFacebookSignOn();
+				});
+			setTimeout(() => attachEventListener(), 0);
```
accbjt: What is the reason for this timeout if it has 0 milliseconds?

accbjt (commented):
@LauraPinilla I left some comments and can we go over how we can test this.

accbjt:
> Just curious but what's reasoning behind adding custom button flag on the custom fields instead of just changing the existing button? Thanks.

@edwardcho1231 I think @LauraPinilla wanted to keep the old version just in case the user wanted the old version. I think we can keep it for now just in case a client wants to use the service providers version of it.

edwardcho1231:
LGTM

accbjt (dismissed):

blocks/identity-block/intl.json
```diff
@@ -2078,5 +2078,26 @@
...
+		"en": "Sign-in prompt was suppressed by the user or dismissed. Please try again later or use another sign-in method."
+	},
+	"identity-block.social-signOn-unlink-error": {
+		"en": "At least one identity needs to remain on the user"
+	},
+	"identity-block.social-signOn-facebook-login": {
+		"en": "Sign in with Facebook"
+	},
+	"identity-block.social-signOn-facebook-signUp": {
+		"en": "Sign up with Facebook"
```
malavikakoppula: @LauraPinilla Can you please add other languages translations as well here?

blocks/identity-block/package.json
```diff
@@ -30,8 +30,12 @@
   },
   "dependencies": {
     "@arc-fusion/prop-types": "^0.1.5",
-    "@arc-publishing/sdk-identity": "^1.79.0",
+    "@arc-publishing/sdk-identity": "^1.88.0",
```
malavikakoppula, line 33: @LauraPinilla Is there any reason why this dependency been added in both block's package.json and outer package.json? In subscriptions-blocks it's still v1.55.0.

malavikakoppula (changes requested):

blocks/identity-block/themes/news.json
```diff
@@ -685,6 +771,64 @@
 			"desktop": {}
 		}
 	},
+	"social-sign-on-button-container-google": {
```
malavikakoppula, line 774: @LauraPinilla Can you please update https://github.com/WPMedia/arc-themes-blocks/blob/ASUB-8521-SocialMediaButtons/.storybook/themes/news.scss as well to match changes in this file.

malavikakoppula (changes requested):

blocks/identity-block/components/social-sign-on/_children/FacebookSignIn.jsx
```diff
@@ -1,20 +1,53 @@
...
+	const phrases = usePhrases();
+	const facebookTextType = socialSignOnIn === SIGN_UP ? "continue_with" : "login_with";
+
+	useEffect(() => {
+		if (customButtons) {
+			document
+				.getElementById("facebook-btn")
+				.addEventListener("click", () => window.onFacebookSignOn());
+		}
+	}, [customButtons]);
```
nschubach: This useEffect that adds an event listener should also remove it when it's cleaned up.  The standard pattern here is:

```
useEffect(() => {
    const listener = () => {}
    element.addEventListener("event", listener)
    return () => element.removeEventListener("event", listener)
}, [dependency]);
```

That return can be inside the condition.

But in this case, you are finding the element that you have created yourself, so why don't you just add the click listener to the Button?  This way you don't need the useEffect.

```
			<Button
				id="facebook-btn"
				variant="secondary-reverse"
				iconLeft={FacebookIcon}
				className={`${className}__Facebook`}
				onClick={() => window?.onFacebookSignOn?.() }
			>
```

blocks/identity-block/components/social-sign-on/_children/GoogleSignIn.jsx
```diff
@@ -1,5 +1,54 @@
...
+						// https://developers.google.com/identity/gsi/web/reference/js-reference
+						// https://developers.google.com/identity/gsi/web/guides/features#exponential_cooldown
+						// eslint-disable-next-line
+						alert(googleNotification);
+						// Remove cookie works in Safari
+						document.cookie = `g_state=;path=/;expires=Thu, 01 Jan 1970 00:00:01 GMT`;
+						window.google.accounts.id.prompt();
+					}
+				}),
+			);
```
nschubach: Again, the same thing here.  Why are we not just adding this to the onClick for the button?

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
```diff
@@ -4,30 +4,59 @@ import { GoogleSignInContext } from "./googleContext";
...
 				if (isOIDC) {
 					loginByOIDC();
-				} else {
-					const validatedURL = validateURL(redirectURL);
-					window.location = validatedURL;
+				} else if (fromSocialSignOnBlock) {
+					Identity.isLoggedIn().then((isLoggedIn) => {
+						if (isLoggedIn) {
+							const validatedURL = validateURL(redirectURL);
+							window.location = validatedURL;
```
nschubach: There's no need to assign it to a constant then immediately set the location to that constant...
```suggestion
							window.location = validateURL(redirectURL);
```

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
```diff
@@ -54,39 +83,96 @@ function useSocialSignIn(redirectURL, isOIDC, socialSignOnIn, onError = () => {}
...
 		};
 		if (config.facebookAppId) {
 			initializeFacebook();
 		}
-	}, [Identity, config.facebookAppId]);
+	}, [Identity, config.facebookAppId, customButtons]);
+
+	useEffect(() => {
+		const checkIsLoggedIn = () => {
+			setTimeout(() => {
```
nschubach: Whenever setting a timeout inside a useEffect, you should have a return to clear the timeout.  useEffects should have a "cleanup".  The reason for this is that a component can be rerendered by it's parent and if there's no cleanup, this timeout can still exist and may get called twice.

A common pattern for this is similar to the addEventListener I noted above:
```
useEffect (() => {
  const timeout = setTimeout(() => { /*... timeout code ...*/ }, 1500)
  return () => clearTimeout(timeout)
}, [dependency])
```

Without that, if a parent component re-renders, the isLoggedIn function may be called multiple times, the promise returned for each of them, the redirect validated multiple times, and the location changed for each of them.  I would not be surprised if this was flagged by something.  Or even worse, not flagged by something if it's trying to get around a security scan...

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
```diff
@@ -54,39 +83,96 @@ function useSocialSignIn(redirectURL, isOIDC, socialSignOnIn, onError = () => {}
...
+				Identity.isLoggedIn().then((isLoggedIn) => {
+					if (isLoggedIn) {
+						const validatedURL = validateURL(redirectURL);
+						window.location = validatedURL;
```
nschubach: Again, a probably needless const.

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
```diff
@@ -54,39 +83,96 @@ function useSocialSignIn(redirectURL, isOIDC, socialSignOnIn, onError = () => {}
...
+			window.FB.Event.subscribe("auth.statusChange", (response) => {
+				if (response.status === "connected") {
+					if (fromSocialSignOnBlock) {
+						checkIsLoggedIn();
+					} else {
+						setIdentityUpdated();
+					}
+				}
+			});
+		}
```
nschubach: This one's complex... you have the timer from above, but you also have this conditional timer that should have a cleanup in the same useEffect.  Is this useEffect doing too much?  Would this be better served as two effects?

blocks/identity-block/features/account-management/_children/SocialSignOnEditableContainer.jsx
```diff
@@ -0,0 +1,98 @@
...
+							// https://developers.google.com/identity/gsi/web/reference/js-reference
+							// https://developers.google.com/identity/gsi/web/guides/features#exponential_cooldown
+							// eslint-disable-next-line
+							alert(googleNotification);
+							// Remove cookie works in Safari
+							document.cookie =  `g_state=;path=/;expires=Thu, 01 Jan 1970 00:00:01 GMT`;
+							window.google.accounts.id.prompt();
+						}
+					}),
+				);
```
nschubach: Would rather see this as an onClick in the button below with a switch for the "Google" type.  Then you would not need the 0 length timer to push this to the end of the call stack as well... see below.

blocks/identity-block/features/account-management/_children/SocialSignOnEditableContainer.jsx
```diff
@@ -0,0 +1,98 @@
...
+			setTimeout(() => attachEventListener(), 0);
+		}
+	}, [type, isConnected, googleNotification]);
+
+	useEffect(() => {
+		if (type === "Facebook" && !isConnected) {
+			const attachEventListener = () =>
+				document.getElementById("custom-Facebook-signin-btn").addEventListener("click", () => {
+					window.onFacebookSignOn();
+				});
```
nschubach: Would rather see this as an onClick in the button below with a switch for the "Facebook" type. Then you would not need the 0 length timer to push this to the end of the call stack as well... see below.

blocks/identity-block/features/account-management/_children/SocialSignOnEditableContainer.jsx
```diff
@@ -0,0 +1,98 @@
...
+			setTimeout(() => attachEventListener(), 0);
+		}
+	}, [type, isConnected]);
+
+	useEffect(() => {
+		if (type === "Apple" && !isConnected) {
+			const attachEventListener = () => {
+				document
+					.getElementById("custom-Apple-signin-btn")
+					.addEventListener("click", () => Identity.initAppleSignOn());
```
nschubach: Would rather see this as an onClick in the button below with a switch for the "Apple" type. Then you would not need the 0 length timer to push this to the end of the call stack as well... see below.

blocks/identity-block/features/account-management/_children/SocialSignOnEditableContainer.jsx
```diff
@@ -0,0 +1,98 @@
...
+					{disconnectText}
+				</Button>
+			)}
+			{!isConnected && (
+				<Button
+					id={`custom-${type}-signin-btn`}
+					className={`${blockClassName}__social-signOn-button-connect`}
+					size="medium"
+					variant="primary"
+				>
```
nschubach: To avoid all the useEffects and DOM querying, just put the onClick here:
```suggestion
					variant="primary"
					onClick={() => {
					  switch(type.toLowerCase()) {
					    case "google": { /* google logic */ }
					    case "facebook": { window.?onFacebookSignOn.?() }
					    case "apple": { Identity.initAppleSignOn() }
					    default : { /* do you need to handle this? */ }
					  }
					}}
				>
```
Or potentially just have that onClick as a [useCallback](https://react.dev/reference/react/useCallback#usecallback) above to keep that logic out of the render if you like.

nschubach (changes requested):
useEffect cleanup requests mainly...

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
LauraPinilla: Hi @nschubach I appreciate all your feedback. 
Regarding this comment.  We'll have two cleanup, but each cleanup will live inside one function, and each function will be called depending on  `fromSocialSignOnBlock` value.  Should we still break the useEffect in two?

blocks/identity-block/intl.json
LauraPinilla: @malavikakoppula for now I added the same English key for all languages, only Irish News is using our blocks, and they are using English. There are few words to create a new order, or we'll need to pay the minimum price per language. 

Screenshot 2024-07-26 at 5 04 04 p m

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
nschubach: You could clear both timers (it doesn't hurt to clear an undefined timer). You could also get rid of the called functions (it looks like they're only being called here) and since it's an event that requires the status change, do we even need the timeout?  Why does it have to wait a second and a half before changing?

nschubach: it doesn't hurt to have multiple effects (it can actually be super helpful to make them laser focused because it doesn't run all the other code every time one or the other changes)

nschubach: Of course, here, it wouldn't make much sense to split those (I missed the dependency earlier) unless those are set up as useCallbacks()

blocks/identity-block/components/social-sign-on/_children/GoogleSignIn.jsx
```diff
@@ -10,31 +10,26 @@ function GoogleSignIn({ customButtons, socialSignOnIn, className }) {
...
-
 	if (customButtons) {
 		return (
 			<Button
 				id="custom-google-signin-btn"
 				variant="secondary-reverse"
 				iconLeft={GoogleIcon}
 				className={`${className}__Google`}
+				onClick={() =>
+					window.google.accounts.id.prompt((notification) => {
```
nschubach: Might want to use conditional calling on this in case "google" does not exist.

`window?.google?.accounts?.id?.prompt`

blocks/identity-block/components/social-sign-on/utils/useSocialSignIn.js
```diff
@@ -135,20 +134,21 @@ function useSocialSignIn(
...
+			const timeout = setTimeout(() => {
 				Identity.isLoggedIn().then((isLoggedIn) => {
 					if (isLoggedIn) {
-						const validatedURL = validateURL(redirectURL);
-						window.location = validatedURL;
+						window.location = validateURL(redirectURL);
 					}
 				});
 			}, 1500);
+  			return () => clearTimeout(timeout)
```
nschubach: This only returns the function to the calling code (159 below) and will not return that to the useEffect...

Since 159/161 are in functions themselves, they will also not return this clean up code to the useEffect.  In this case, you need a shared `timeout` pointer defined at the top of the useEffect, which you clean up at the end of the useEffect.
```
useEffect(() => {
  let timeout;
  const checkIsLoggedIn ...
    timeout = setTimeout(...
  };
  const setIdentityUpdated ...
    timeout = setTimeout(...
  };
  ...
  // rest of code
  ...
  return () => clearTimeout(timeout);
}, [ ... ]);
```
Make sense?

nschubach: this should be okay since only one of those timeouts will occur.

nschubach (approved):
Definitely feel better about the cleanup, thanks Laura!

malavikakoppula (approved):

PR #2350, opened 2025-08-08
remove twitter embed, only render blockquote to improve chromatic con…

annasherman:
…sistency

## Description

This will set a new baseline for chromatic builds (once accepted) that will hopefully remove some of the twitter oembed inconsistencies that we see in chromatic. This should be acceptable for a storybook mock and prevent us from seeing false positive. The only real way to test is just to merge and see if subsequent changes trigger these.


annasherman:
also a weird package.json formatting fix in here

DavidAZimmerly (approved):

annasherman:
approved changes as new baseline

